PR #85, opened 2020-11-03
New services, update readme

### LINKED ISSUE #64: Backend: Scripts for nuking DB

Christdej:
Closes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,17 @@
 # Barrier Management Tool
 [![Build Status](https://dev.azure.com/lambdaville/Fusion-BMT/_apis/build/status/equinor.fusion-bmt?branchName=master)](https://dev.azure.com/lambdaville/Fusion-BMT/_build/latest?definitionId=22&branchName=master)
 
-### Backend configuration
+## Backend configuration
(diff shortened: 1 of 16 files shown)


backend/tests/AnswerService.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        [Fact]
+        public void Create()
+        {
+            AnswerService answerService = new AnswerService(_context);
+
+            int nAnswerBefore = answerService.GetAll().Count();
+            Answer answer = answerService.Create(ExampleParticipant(), Progression.Alignment, ExampleQuestion(), Severity.High, "test_answer");
+            int nanswersAfter = answerService.GetAll().Count();
```
nicholasdalhaug, line 38: nAnswersAfter with correct capitalization

nicholasdalhaug (approved):
I believe there are some codacy info and warnings here we might want to fix. And I had one comment. But it looks good!

backend/tests/AnswerService.cs
Christdej: yeah, thank you =)

PR #86, opened 2020-11-03
Frontend: Create Question and answer component

nicholasdalhaug:


ReedOnly (approved):

frontend/src/App.tsx
```diff
@@ -19,6 +47,16 @@ const App = () => {
...
+            <div style={{border: "10px solid black"}}>
+                <QuestionAndAnswerForm
+                    questionNumber={1}
+                    question={question}
+                    answer={answer}
+                    onAnswerChange={(answer) => {
+                        setAnswer(answer);
+                    }}
+                />
+            </div>
```
tmals-equi, line 59: Do we need to have this "demo" in App?

tmals-equi (commented):
Only one question, looks good otherwise :)

frontend/src/App.tsx
nicholasdalhaug: Nah, you are probably right. Will remove it.

PR #93, opened 2020-11-05
Delete participant

Christdej:


nicholasdalhaug (approved):
I see this is a new service method. Great having a test for this! I assume that we will need a mutation later, and that the mutation does the checking of if the stage we are at is Nomination or not and only allow deletion if it is.

PR #101, opened 2020-11-09
Use backend in creating evaluation

nicholasdalhaug:


frontend/src/views/Project/Dashboard/CreateEvaluationButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+interface CreateEvaluationMutationProps {
+    createEvaluation: (azureUniqueId: string, name: string, projectId: string) => void
+    loading: boolean
+    evaluation: Evaluation | undefined
+    error: ApolloError | undefined
+}
+
+const useCreateEvaluationMutation = (): CreateEvaluationMutationProps => {
+    const ADD_PROJECT = gql`
+        mutation CreateEvaluation($azureUniqueId: String!, $name: String!, $projectId: String!){
```
tmals-equi, line 19: Our backend models use fusion IDs rather than azure unique IDs. I realize they are the same, but it might get confusing if we keep using both types of names?

frontend/src/views/Project/Dashboard/CreateEvaluationDialog.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from 'react'
+
+import { Button, ModalSideSheet, TextInput } from '@equinor/fusion-components'
+import { Typography } from '@equinor/eds-core-react'
+import { Container, Grid } from '@material-ui/core'
+
+const NAME_MIN_LENGTH = 3
```
tmals-equi, line 7: Why 3? Could it be enough to just demand that the string is not empty?

frontend/src/views/Project/Dashboard/CreateEvaluationButton.tsx
nicholasdalhaug: I totally agree that we should not use both names! I think we recently changed the names from `fusionUserId` to `azureUniqueId`.

frontend/src/views/Project/Dashboard/CreateEvaluationDialog.tsx
nicholasdalhaug: I guess you can say it is an arbitrary number greater than zero and small. I agree though, we can just ensure that it is not empty.

tmals-equi (approved):

PR #104, opened 2020-11-09
Preparation view component

tmals-equi:


frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
+
+import * as React from 'react';
+import { Chip, NavigationStructure, NavigationDrawer } from '@equinor/fusion-components';
+import { Barrier, Answer, Evaluation, Question, } from '../api/models';
+import QuestionAndAnswerForm from '../components/QuestionAndAnswer/QuestionAndAnswerForm';
+import { Box } from '@material-ui/core';
+
+interface BarrierQuestionsProps
+{
+    title: string
```
nicholasdalhaug, line 10:
```suggestion
    barrier: Barrier
```

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+    questions: Question[]
+}
+
+const BarrierQuestions = ({title, questions}: BarrierQuestionsProps) => {
```
nicholasdalhaug, line 14: Could we rather call this something like 

```suggestion
const BarrierQuestionsView = ({barrier, questions}: BarrierQuestionsViewProps) => {
```

Or `Form` if that is more appropriate?

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return (
+        <>
+            <h2>{title}</h2>
```
nicholasdalhaug, line 17: Should we rather use eds `Typography`here?

```suggestion
            <Typography variant="h2">{title}</Typography>
```

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+            {questions.map((question) => {
+                return (
+                    <>
+                        <hr style={{ height: "1", color: "lightgray" }}/>
```
nicholasdalhaug, line 21: Is there no divider in Fusion or EDS to use instead?

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+    questions: Question[]
+}
+
+const BarrierQuestions = ({title, questions}: BarrierQuestionsProps) => {
+    return (
+        <>
+            <h2>{title}</h2>
+            {questions.map((question) => {
+                return (
+                    <>
```
nicholasdalhaug, line 20: Every dynamically generated list should have a key prop in react

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+                        <hr style={{ height: "1", color: "lightgray" }}/>
+                        <QuestionAndAnswerForm
+                            question={question}
+                            answer={question.answers[0]}
```
nicholasdalhaug, line 24: There should be some filtering here to only view the answers from the user, if it is not guaranteed from the backend. By this I mean that we should remove the magic indexing `[0]` in some way.

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+                            questionNumber={1}
```
nicholasdalhaug, line 25: The question number should probably be generated dynamically.

frontend/src/views/PreparationView.tsx
nicholasdalhaug: I also think that this could be moved to its own file.

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+const PreparationView = ({evaluation}: PreparationViewProps) => {
+    const [selectedBarrier, setSelectedBarrier] = React.useState<string>(Object.values(Barrier)[0]);
+    const [structure, setStructure] = React.useState<NavigationStructure[]>(
+        Object.entries(Barrier).map(([shortName, barrierName]) => {
+            return {
+                id: barrierName,
+                type: 'grouping',
+                title: barrierName,
+                icon: <>{shortName}</>,
+                aside: <Chip title="3/10" />
```
nicholasdalhaug, line 52: This number should be generated from the questions and answers

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <Box flexGrow={0}>
+                <NavigationDrawer
+                    id="navigation-drawer-story"
+                    structure={structure}
+                    selectedId={selectedBarrier}
+                    onChangeSelectedId={(selectedItem) => setSelectedBarrier(selectedItem)}
+                    onChangeStructure={(newStructure) => {
+                        setStructure(newStructure);
+                    }}
+                />
```
nicholasdalhaug, line 68: Can we move this together with some of the code above into its own component and even its own file?

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+            </Box>
+            <Box>
+                <div style={{margin: 20}}>
```
nicholasdalhaug, line 71: We could do 
```suggestion
            <Box m={20}>
```
here, but how about we just stick to one method and use the MUI `Container`?

frontend/src/views/PreparationView.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    <BarrierQuestions title={selectedBarrier} questions={evaluation.questions.filter(q => q.barrier === selectedBarrier)} />
```
nicholasdalhaug, line 72: I think we should have some nicer formatting of this long line. 

```suggestion
                    <BarrierQuestions 
                        title={selectedBarrier} 
                        questions={
                            evaluation.questions.filter(
                                q => q.barrier === selectedBarrier
                            )
                        } 
                    />
```
It does not need to be this split up, but maybe more than it is?

nicholasdalhaug (changes requested):
I think some of my comments here are important for not making the integration with the backend a really big job. I also see some Codacy messages that we might want fixed.

frontend/src/views/PreparationView.tsx
tmals-equi: Hmm, I'm not sure how you reproduce this behaviour with Container?

nicholasdalhaug (approved):
Looks very good to me!

PR #105, opened 2020-11-09
Question Template

### LINKED ISSUE #74: Backend: Create QuestionService

### LINKED ISSUE #75: Backend: Create QuestionsTemplateService

### LINKED ISSUE #76: Backend: Script for populating SQL db with questions

Christdej:
Closes #75 
Closes #74 
Closes #76

### DIFF
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -1,5 +1,6 @@
 using api.Services;
 using api.Models;
+using System.Collections.Generic;
 
 namespace api.GQL
 {
@@ -10,13 +11,15 @@ namespace api.GQL
         private readonly ParticipantService _participantService;
         private readonly QuestionService _questionService;
         private readonly AnswerService _answerService;
+        private readonly QuestionTemplateService _questionTemplateService;
 
         public Mutation(
             ProjectService projectService,
             EvaluationService evaluationService,
             ParticipantService participantService,
             QuestionService questionService,
-            AnswerService answerService
+            AnswerService answerService,
+            QuestionTemplateService questionTemplateService
         )
         {
             _projectService = projectService;
@@ -24,18 +27,28 @@ namespace api.GQL
             _participantService = participantService;
             _questionService = questionService;
             _answerService = answerService;
+            _questionTemplateService = questionTemplateService;
         }
 
-        public Evaluation CreateEvaluation(string name, string projectId, Participant creator)
+        // TODO: This methods should not take azureUniqueId as parameter but get it from token
+        public Evaluation CreateEvaluation(string name, string projectId, string azureUniqueId)
         {
             Project project = _projectService.GetProject(projectId);
-            return _evaluationService.Create(name, project, creator);
+            Evaluation evaluation = _evaluationService.Create(name, project);
+            _participantService.Create(azureUniqueId, evaluation, Organization.All, Role.Facilitator);
+
+            List<QuestionTemplate> questionTemplates = _questionTemplateService.ActiveQuestions();
+            foreach (QuestionTemplate template in questionTemplates)
+            {
+                _questionService.Create(template, evaluation);
+            }
+            return evaluation;
         }
 
-        public Participant CreateParticipant(string fusionPersonId, string evaluationId, Organization organization, Role role)
+        public Participant CreateParticipant(string azureUniqueId, string evaluationId, Organization organization, Role role)
         {
             Evaluation evaluation = _evaluationService.GetEvaluation(evaluationId);
-            return _participantService.Create(fusionPersonId, evaluation, organization, role);
+            return _participantService.Create(azureUniqueId, evaluation, organization, role);
         }
 
         public void DeleteParticipant(string participantId)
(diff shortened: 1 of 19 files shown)


backend/api/Services/QuestionTemplateService.cs
```diff
@@ -0,0 +1,66 @@
...
+        public List<QuestionTemplate> ActiveQuestions()
+        {
+            List<QuestionTemplate> questions = new List<QuestionTemplate>();
+            foreach (QuestionTemplate template in _context.QuestionTemplates)
+            {
+                if (template.Status.Equals(Status.Active))
+                {
+                    questions.Add(template);
+                }
+            }
```
nicholasdalhaug, line 30: I think it might be better in the long run to use the database for filtering. We should maybe avoid for each loops where we can. 

```suggestion
            List<QuestionTemplate> questions = _context.QuestionTemplates.Where(template => template.Status.Equals(Status.Active));
```

backend/tests/Services/QuestionTemplateService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        [Fact]
+        public void GetQueryable()
+        {
+            QuestionTemplateService questionTemplateService = new QuestionTemplateService(_context);
+
+            IQueryable<QuestionTemplate> questions = questionTemplateService.GetAll();
+
+            Assert.True(questions.Count() > 0);
+        }
```
nicholasdalhaug, line 27: We could also have tests for getAll and getActive here. And maybe others(?), this was just with a quick glance.

nicholasdalhaug (changes requested):
Nice with new functionality! I have some suggestions here for code, also some suggested tests and I think some of the Codacy messages had a point, so please look at them as well. If any of the Codacy messages are wrong, I think we should fix the linting so that we do not have to see the wrong messages.

backend/api/GQL/Mutation.cs
```diff
@@ -10,32 +11,45 @@ public class Mutation
...
+            _questionTemplateService = questionTemplateService;
         }
 
-        public Evaluation CreateEvaluation(string name, string projectId, Participant creator)
+        // TODO: This methods should not take azureUniqueId as parameter but get it from token
+        public Evaluation CreateEvaluation(string name, string projectId, string azureUniqueId)
         {
             Project project = _projectService.GetProject(projectId);
-            return _evaluationService.Create(name, project, creator);
+            Evaluation evaluation = _evaluationService.Create(name, project);
```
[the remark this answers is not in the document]
nicholasdalhaug: An evaluation is required in order to create a participant or a question. Case resolved 😄

backend/api/Services/QuestionTemplateService.cs
Christdej: yes agreed

backend/tests/Services/QuestionTemplateService.cs
Christdej: Ill make a test for getActive aswell, the Create() method is private for now, as it is not supposed to be used yet

backend/tests/Services/QuestionTemplateService.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        [Fact]
+        public void ActiveQuestions()
+        {
+            QuestionTemplateService questionTemplateService = new QuestionTemplateService(_context);
+
+            List<QuestionTemplate> questionTemplates = questionTemplateService.ActiveQuestions();
+
+            Assert.True(questionTemplates.Count > 0);
```
nicholasdalhaug, line 37: Should we also create a question template with an inactive status and see that it does not get fetched?

backend/tests/Services/QuestionTemplateService.cs
nicholasdalhaug: OK. Then we need to remember to test it when it gets used!

nicholasdalhaug (changes requested):

backend/api/GQL/Mutation.cs
```diff
@@ -10,32 +11,45 @@ public class Mutation
...
+            _participantService.Create(azureUniqueId, evaluation, Organization.All, Role.Facilitator);
+
+            List<QuestionTemplate> questionTemplates = _questionTemplateService.ActiveQuestions();
+            List<Question> questions = new List<Question>();
+            foreach (QuestionTemplate template in questionTemplates)
+            {
+                questions.Add(_questionService.Create(template, evaluation));
+            }
```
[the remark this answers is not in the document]
Christdej: Good point! Will fix

backend/tests/Services/QuestionTemplateService.cs
Christdej: Then I would have to make the Create QuestionTemplate method public and test that aswell, which we will probably do at a later point when QuestionTemplates are supposed to be editable. So maybe postpone this till then?

nicholasdalhaug: Okay. I'll make it an issue.

nicholasdalhaug (approved):

ReedOnly (approved):

PR #109, opened 2020-11-09
Add .editorconfig rules and refactor switch statement

ReedOnly:


nicholasdalhaug (approved):
Nice! I did not know the switch could be that simple.

PR #115, opened 2020-11-10
Add graphql-codegen for generating ts models

### LINKED ISSUE #103: Frontend: Generate models.ts from grapql schema

ReedOnly:
closes #103


backend/api/Models/Models.cs
```diff
@@ -159,7 +160,28 @@ public enum Progression
 
     public enum Barrier
     {
-        GM, PS1, PS2, PS3, PS4, PS5, PS6, PS7, PS12, PS15, PS22
+        [Description("General Matters")]
```
nicholasdalhaug, line 163: Are these descriptions doing anything?

doc/schema.graphql
```diff
@@ -0,0 +1,276 @@
+schema {
```
nicholasdalhaug, line 1: Shouldn't this file be autogenerated? And if so, should it be removed from the repo?

frontend/schema.yml
```diff
@@ -0,0 +1,10 @@
+overwrite: true
+
+# schema: ../doc/schema.graphql
```
nicholasdalhaug, line 3: I see the value of this comment for debugging. Maybe we should consider adding it to the readme instead?

frontend/src/api/models.ts
```diff
@@ -1,149 +1,308 @@
...
+  organization: Organization;
+  text?: Maybe<Scalars['String']>;
+  supportNotes?: Maybe<Scalars['String']>;
+  barrier: Barrier;
+  createDate: Scalars['DateTime'];
+  answers?: Maybe<Array<Maybe<AnswerInput>>>;
+  actions?: Maybe<Array<Maybe<ActionInput>>>;
+  evaluation?: Maybe<EvaluationInput>;
+  questionTemplate?: Maybe<QuestionTemplateInput>;
+};
```
nicholasdalhaug, line 234: The Input types such as these I think are redundant. I do not think they are used, maybe we should get rid of them?There are also some other types here that are not used and we do not want to maintain.

frontend/src/api/modelsCustom.ts
```diff
@@ -0,0 +1,149 @@
+export type Action = {
```
nicholasdalhaug, line 1: I guess we can remove this file as well? What is it for?

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerForm.tsx
```diff
@@ -52,7 +52,7 @@ const QuestionAndAnswerForm = ({questionNumber, question, answer, onAnswerChange
                 <Box display="flex">
                     <Box mr={5}>
                         <AnswerSeverityForm
-                            severity={answer.severity}
+                            severity={answer.severity || Severity.Na}
```
nicholasdalhaug, line 55: I think we should rather enforce severity to be set than to check for undefined everywhere.

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerForm.tsx
```diff
@@ -17,7 +17,7 @@ interface QuestionAndAnswerFormProps {
 }
 
 const QuestionAndAnswerForm = ({questionNumber, question, answer, onAnswerChange}: QuestionAndAnswerFormProps) => {
-    const [markdown, setMarkdown] = useState<string>(answer.text);
+    const [markdown, setMarkdown] = useState<string>(answer.text || "");
```
nicholasdalhaug, line 20: Same as below. We should enforce it to be defined.

nicholasdalhaug (changes requested):

doc/schema.graphql
ReedOnly: Yes, it was until I had the file from server working

backend/api/Models/Models.cs
ReedOnly: No, we can remove

frontend/schema.yml
ReedOnly: think for now we can just remove

frontend/src/api/modelsCustom.ts
ReedOnly: The old manual file. Will remove

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerForm.tsx
ReedOnly: Will fix in backend

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerForm.tsx
ReedOnly: yes

frontend/package.json
```diff
@@ -6,7 +6,8 @@
     "scripts": {
         "start": "fusion start-app",
         "build": "fusion build-app",
-        "test": "jest"
+        "test": "jest",
+        "schema": "graphql-codegen --config schema.yml"
```
nicholasdalhaug, line 10: Could this file be moved to api, next to the models.ts?

frontend/src/api/models.ts
```diff
@@ -1,149 +1,300 @@
...
-    PS1 = "Containment",
-    PS2 = "HVAC",
-    PS3 = "Leak Detection",
-    PS4 = "ESD",
-    PS6 = "Ignition Source Control",
-    PS7 = "Fire Detection",
-    PS12 = "Process Safety",
-    PS15 = "Layout",
-    PS22 = "HMI"
-}
```
nicholasdalhaug, line 38: I think we should fix everywhere that this is used before merging to master. This we change the barrier dialog/sidesheet. Or at least make an issue on the board for fixing it with the highest priority.

frontend/schema.yml
```diff
@@ -0,0 +1,10 @@
+overwrite: true
+
+schema:
+  - http://localhost:5000/graphql?sdl:
+generates:
+  ./src/api/models.ts:
+    plugins:
+      - "typescript"
+    config:
+      maybeValue: T
```
nicholasdalhaug, line 10: Is it possible to have a comment here about what this does?

frontend/src/api/models.ts
nicholasdalhaug: If we cannot remove these I really hope we do not get much problems from the typescript definitions. It can be difficult to get the correct typescript when it is based on C# and has to go through 2 layers of transpiling.

frontend/src/api/models.ts
ReedOnly: Will make an Issue to the board

frontend/schema.yml
ReedOnly: done

nicholasdalhaug (approved):

PR #118, opened 2020-11-11
Evaluation list

tmals-equi:


backend/api/Services/ProjectService.cs
```diff
@@ -19,7 +20,9 @@ public ProjectService([Service] BmtDbContext context)
 
         public Project EnsureCreated(string fusionProjectID)
         {
-            Project project = _context.Projects.FirstOrDefault(project => project.FusionProjectId.Equals(fusionProjectID));
+            Project project = _context.Projects
+                .Include(project => project.Evaluations)
```
nicholasdalhaug, line 24: This should not be necessary anymore, right?

frontend/src/views/Project/Dashboard/EvaluationListItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const EvaluationListItem = ({evaluation}: EvaluationListItemProps) => {
+    const resolveIconColor = (progression: Progression) => {
+        const keys = Object.keys(Progression)
+        const values = Object.values(Progression)
+        if (keys.indexOf(evaluation.progression) < values.indexOf(progression)) return "lightgray"
+        if (keys.indexOf(evaluation.progression) > values.indexOf(progression)) return "green"
+        return "orange"
+    }
```
nicholasdalhaug, line 19: This function should be made simpler. Maybe some reusability as well. How about a function from currentProgression and progressionToCheck called `checkProgressionStatus` which returns a progression status of `In Progress`, `Completed` or `Awaiting`? That function is then testable as well. Then simply use that function in this component.

frontend/src/views/Project/Dashboard/EvaluationListItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Link to={`${pathname}/evaluation/${evaluation.id}`} style={{textDecoration: "none"}}>
+            <Box display="flex" flexDirection="row" alignItems="center" marginY="0.7rem">
+                <Box width="200px"><Typography variant="body_short" link token={{
+                    fontSize: '1.2rem',
+                }}>{evaluation.name}</Typography></Box>
+                <CheckCircleIcon
+                    color={resolveIconColor(Progression.Nomination)}
+                    height={24}
+                    width={24}
+                />
```
nicholasdalhaug, line 32: I think we should make a new component here that itself figures out if the symbol should be a check of a clock, like in the Figma sketches. And maybe also finds out its own color depending on the input.

frontend/src/views/Project/Dashboard/ProjectDashboardGQL.ts
```diff
@@ -0,0 +1,93 @@
...
+                        evaluations(existingEvaluations = []) {
+                            const newEvaluationRef = cache.writeFragment({
+                                id: createEvaluation.id,
+                                data: createEvaluation,
+                                fragment: gql`
+                                fragment NewEvaluation on Evaluation {
+                                    id
+                                    createDate
+                                    name
+                                    projectId
```
nicholasdalhaug, line 71: Have you checked that his works and that you do not need to use the same fragment as in the query for the caching to work?

frontend/src/views/Project/Dashboard/ProjectDashboardGQL.ts
```diff
@@ -0,0 +1,93 @@
...
+        mutation CreateEvaluation($azureUniqueId: String!, $name: String!, $projectId: String!){
+            createEvaluation(
+                azureUniqueId: $azureUniqueId
+                name: $name
+                projectId: $projectId
+            ){
+                id
+                createDate
+                name
+                projectId
```
nicholasdalhaug, line 52: You can also reuse the newly created fragment here.

frontend/src/views/Project/Dashboard/ProjectDashboardGQL.ts
```diff
@@ -0,0 +1,93 @@
...
+interface EvaluationQueryProps {
+    loading: boolean
+    evaluations: Evaluation[] | undefined
+    error: ApolloError | undefined
+}
+
+export const useEvaluationsQuery = (projectId: string): EvaluationQueryProps => {
+    const GET_EVALUATIONS = gql`
+        query {
+            evaluations(where: {projectId: {eq: "${projectId}"}}) {
```
nicholasdalhaug, line 14: I think that using projectId like this is not ideal. We should rather do queries like GraphQL suggests by referring to the property `project`. See `Nesting your objects in queries` [here](https://atheros.ai/blog/graphql-best-practices-for-graphql-schema-design).

frontend/src/views/Project/Dashboard/ProjectDashboardView.tsx
```diff
@@ -1,18 +1,69 @@
...
+    if (loadingQuery){
+        return <>
+            Loading...
+        </>
+    }
+
+    if (errorQuery !== undefined) {
+        return <div>
+            <TextArea
+                value={`Error in loading participants: ${JSON.stringify(errorQuery)}`}
```
nicholasdalhaug, line 27: I think there might be some copy-paste error here and same below.

nicholasdalhaug (changes requested):

backend/api/Services/ProjectService.cs
tmals-equi: Why not?

nicholasdalhaug: Because you are not fetching the evaluation by getting the evaluation under project. You  should get the evaluation from the evaluations query and filter on the underlying project.

tmals-equi: Yeah true, but do we know we'll never do it like that?

frontend/src/views/Project/Dashboard/ProjectDashboardGQL.ts
nicholasdalhaug:
```suggestion
            evaluations(where: {project: {id: {eq: "${projectId}"}}}) {
```

backend/api/Services/ProjectService.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Linq;
+using Microsoft.EntityFrameworkCore;
```
nicholasdalhaug, line 3: I think you can remove this import.

frontend/src/routes/ProjectRoute.tsx
```diff
@@ -16,9 +16,9 @@ const useProjectQuery = (fusionProjectId: string): ProjectQueryProps => {
     const GET_PROJECT = gql`
         query {
             project(fusionProjectID: "${fusionProjectId}") {
-                id
+                id,
```
nicholasdalhaug, line 19: Maybe we do not need commas here?

nicholasdalhaug (approved):

PR #126, opened 2020-11-16
DB context tests

### LINKED ISSUE #107: Backend: Test for BmtDbContext model connections

Christdej:
Closes #107

### DIFF
--- a/backend/tests/DbContext.cs
+++ b/backend/tests/DbContext.cs
@@ -1,20 +1,74 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Xunit;
 
 using api.Context;
+using api.Services;
 using api.Models;
 
 namespace tests
 {
+    [Collection("UsesDbContext")]
     public class DbContextTests
     {
+        private readonly BmtDbContext _context;
+        public DbContextTests()
+        {
+            _context = Globals.context;
+        }
+
         [Fact]
         public void InitQuestions()
         {
             List<Question> questions = InitContent.Questions;
 
             Assert.Equal(11, questions.Count());
+        }
+
+
+        [Fact]
+        public void ParticipantAddedToEvaluation()
+        {
+            EvaluationService evaluationService = new EvaluationService(_context);
+            Evaluation exampleEvaluation = evaluationService.GetAll().First();
+            ParticipantService participantService = new ParticipantService(_context);
+
+            int participantsBefore = exampleEvaluation.Participants.Count;
+            participantService.Create("some_id_we_got_from_fusion", exampleEvaluation, Organization.Engineering, Role.Participant);
+            int participantsAfter = exampleEvaluation.Participants.Count;
+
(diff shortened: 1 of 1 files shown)


backend/tests/DbContext.cs
```diff
@@ -1,20 +1,75 @@
...
+        {
+            EvaluationService evaluationService = new EvaluationService(_context);
+            Evaluation exampleEvaluation = evaluationService.GetAll().First();
+            int participantsBefore = exampleEvaluation.Participants.Count;
+            ParticipantService participantService = new ParticipantService(_context);
+
+            participantService.Create("some_id_we_got_from_fusion", exampleEvaluation, Organization.Engineering, Role.Participant);
+            int participantsAfter = exampleEvaluation.Participants.Count;
+
+            Assert.True(participantsBefore + 1 == participantsAfter);
```
nicholasdalhaug, line 41: We get a better output from testing if we use `Assert.Equals` instead. Same farther down.

backend/tests/DbContext.cs
```diff
@@ -1,20 +1,75 @@
...
             Assert.Equal(11, questions.Count());
         }
+
+
+        [Fact]
+        public void ParticipantAddedToEvaluation()
+        {
+            EvaluationService evaluationService = new EvaluationService(_context);
+            Evaluation exampleEvaluation = evaluationService.GetAll().First();
+            int participantsBefore = exampleEvaluation.Participants.Count;
```
nicholasdalhaug, line 35: I think a newline right here, to follow the AAA(Arrange, Act, Assert) pattern would be nice.

backend/tests/DbContext.cs
```diff
@@ -1,20 +1,75 @@
...
+
+        [Fact]
+        public void ParticipantAddedToProject()
+        {
+            ProjectService projectService = new ProjectService(_context);
+            Project exampleProject = projectService.GetAll().First();
+            Evaluation exampleEvaluation = exampleProject.Evaluations.First();
+
+            int participantsBefore = exampleProject.Evaluations.First().Participants.Count;
+            ParticipantService participantService = new ParticipantService(_context);
```
nicholasdalhaug, line 52: The participant service could be initialized before the int, thereby also using the AAA(Arrange, Act, Assert) pattern.

nicholasdalhaug (changes requested):

backend/tests/DbContext.cs
Christdej: Good point, will fix =)

backend/tests/DbContext.cs
Christdej: Done

backend/tests/DbContext.cs
Christdej: Done

ReedOnly (approved):

nicholasdalhaug (approved):

PR #129, opened 2020-11-17
Show questions from backend

nicholasdalhaug:


frontend/src/components/QuestionAndAnswer/QuestionAndAnswerFormWithApi.tsx
```diff
@@ -0,0 +1,91 @@
...
+    createDate: "",
+    answeredBy: emptyParticipant
+}
+
+interface QuestionAndAnswerFormWithApiProps {
+    questionNumber: number
+    question: Question
+}
+
+const QuestionAndAnswerFormWithApi = ({questionNumber, question}: QuestionAndAnswerFormWithApiProps) => {
```
tmals-equi, line 64: Not sure about this name, could we change it to something like QuestionAndAnswerFormWithData?

tmals-equi (approved):
Looks good apart from one minor question

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerFormWithApi.tsx
nicholasdalhaug: Doesn't that make it sound like it needs more data as input, when the case is that it needs less? I was also thinking that this whole thing should not be necessary and that we should get big queries to work correctly with the cache. What do you think about that?

tmals-equi: Yeah, I agree. But maybe that could be a separate issue?

nicholasdalhaug: You are probably right.

PR #130, opened 2020-11-17
Remove cascade deletion on answer

Christdej:
Creating initial database migration complained because Answer had multiple cascade paths, so removing cascading from answer solves problem.

### DIFF
--- a/backend/api/Context/BmtDbContext.cs
+++ b/backend/api/Context/BmtDbContext.cs
@@ -77,7 +77,8 @@ namespace api.Context
             modelBuilder.Entity<Question>()
                 .HasMany(t => t.Answers)
                 .WithOne(t => t.Question)
-                .HasForeignKey(t => t.QuestionId);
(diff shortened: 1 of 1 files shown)


backend/api/Context/BmtDbContext.cs
```diff
@@ -87,7 +88,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<QuestionTemplate>()
                 .HasMany(t => t.Questions)
                 .WithOne(t => t.QuestionTemplate)
-                .HasForeignKey(t => t.QuestionTemplateId);
+                .HasForeignKey(t => t.QuestionTemplateId)
+                .OnDelete(DeleteBehavior.NoAction);
```
nicholasdalhaug, line 92: Will question template still be deleted if the migration requires it?

nicholasdalhaug (approved):

PR #142, opened 2020-11-18
Delete nominee button

tmals-equi:


frontend/src/views/Evaluation/Nomination/NominationTable.tsx
```diff
@@ -16,9 +20,23 @@ interface DataTableRowProps {
 }
 
 const DataTableItemRenderer: React.FC<DataTableRowProps> = ({ item }) => (
-    <ParticipantCard participant={item.participant} />
+    <ParticipantCard key={item.participant.id} participant={item.participant} />
```
nicholasdalhaug, line 23: We. do not need this anymore, right? We just tested it. There was an other solution.

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
```diff
@@ -16,9 +20,23 @@ interface DataTableRowProps {
...
 )
 
+const DeleteColumn: React.FC<DataTableRowProps> = ({ item }) => {
```
nicholasdalhaug, line 26: Could we call this something like (?)
```suggestion
const DeleteColumnItemRenderer: React.FC<DataTableRowProps> = ({ item }) => {
```

We should anyways call the two renderers something similar.

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
```diff
@@ -16,9 +20,23 @@ interface DataTableRowProps {
...
+    if (item.role === Role.Facilitator) return null
```
nicholasdalhaug, line 27: Could we return a react fragment instead? 
```suggestion
    if (item.role === Role.Facilitator) return <></>
```

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
```diff
@@ -16,9 +20,23 @@ interface DataTableRowProps {
...
+    const {deleteParticipant, loading} = useDeleteParticipantMutation()
+
+    return (
+        <Button onClick={() => {
+            deleteParticipant(item.participant.id)
+        }}
+        disabled={item.progressionStatus !== ProgressionStatus.InProgress || loading}
+        >Delete</Button>
+    )
+}
```
nicholasdalhaug, line 38: Could the content of this be moved to a new file or something. Especially since this does a call to the backend, I think it might be nice to not hide that inside this complex file for DataTable abstraction.

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
```diff
@@ -39,24 +57,34 @@ const columns: DataTableColumn<DataTableItem>[] = [
...
+        accessor: 'participant',
+        label: '',
+        component: DeleteColumn,
+        priority: 3,
+    },
 ]
 
 interface NominationTableProps {
     participants: Participant[]
+    progressionStatus: ProgressionStatus
```
nicholasdalhaug, line 71: Is it better to call it (?)
```suggestion
    currentProgressionStatus: ProgressionStatus
```

nicholasdalhaug (changes requested):
Nice! I just have a few comments.

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
tmals-equi: I'm not sure I think that's necessary for such a small component. Also, then we'd need to either define the DataTableRowProps twice, since both DeleteColumnItemRenderer and DataTableItemRenderer use it

nicholasdalhaug (approved):

PR #149, opened 2020-11-23
Enum to string

tmals-equi:


frontend/src/utils/EnumToString.ts
```diff
@@ -0,0 +1,63 @@
...
+    case Barrier.Ps7:
+        return "Fire Detection"
+    case Barrier.Ps12:
+        return "Process Safety"
+    case Barrier.Ps15:
+        return "Layout"
+    case Barrier.Ps22:
+        return "HMI"
+    default:
+        return ""
```
nicholasdalhaug, line 29: Why do we need default?

```suggestion
```

frontend/src/utils/EnumToString.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+let OrganizationToString = (organization: Organization): string => {
```
nicholasdalhaug, line 33: Same as above: const , lower case and no default? Same below as well.

nicholasdalhaug (changes requested):

ReedOnly (approved):

PR #152, opened 2020-11-24
Preventing adding already nominated participant

tmals-equi:


nicholasdalhaug (commented):
Looks good, but should we fix backend for this as well?

tmals-equi:
> Looks good, but should we fix backend for this as well?

Yeah I think so, but maybe that's a separate issue?

nicholasdalhaug (approved):

PR #158, opened 2020-11-25
Alignment view and answer summary

tmals-equi:


frontend/src/components/BarrierQuestionsView.tsx
```diff
@@ -3,20 +3,22 @@ import * as React from 'react'
...
+import { barrierToString } from '../utils/EnumToString'
+import QuestionAndAnswerFormWithApi from './QuestionAndAnswer/QuestionAndAnswerFormWithApi'
+import { getAzureUniqueId } from '../utils/Variables'
 
 interface BarrierQuestionsViewProps
 {
     barrier: Barrier
     questions: Question[]
+    showAnswerSummaryButton: boolean
+    onQuestionSummarySelected: (question: Question) => void
```
nicholasdalhaug, line 16: Maybe remove the redundant prop?

```suggestion
    onQuestionSummarySelected?: (question: Question) => void
```

frontend/src/components/AnswerSummarySidebar.tsx
```diff
@@ -0,0 +1,40 @@
...
+import SingleAnswerSummary from './SingleAnswerSummary'
+
+interface AnswerSummarySidebarProps
+{
+    open: boolean
+    onCloseClick: () => void
+    question: Question
+}
+
+const AnswerSummarySidebar = ({ open, onCloseClick, question }: AnswerSummarySidebarProps) => {
```
nicholasdalhaug, line 15: I think it needs to know what progression we are on. We do not want the preparation answers to come up in the workshop.

frontend/src/components/BarrierQuestionsView.tsx
```diff
@@ -30,6 +32,8 @@ const BarrierQuestionsView = ({barrier, questions}: BarrierQuestionsViewProps) =
                             questionNumber={idx + 1}
                             question={question}
                             answer={answer}
+                            showAnswerSummaryButton={ showAnswerSummaryButton }
+                            onQuestionSummarySelected={onQuestionSummarySelected}
```
nicholasdalhaug, line 36: I do not like that we have to pass props a long way down, but I am not sure how we can fix it either. Could some context providers be helpful? Or maybe creating new components instead of changing the ones that exist.

frontend/src/components/SeverityIndicator.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react'
+import { RadioButton } from '@equinor/fusion-components'
+
+interface SeverityIndicatorProps
+{
+    severity: Severity
+}
+
+const SeverityIndicator = ({severity}: SeverityIndicatorProps) => {
+    let getColor = (severity: Severity) => {
```
nicholasdalhaug, line 11:
```suggestion
    const getColor = (severity: Severity) => {
```

frontend/src/components/SingleAnswerSummary.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Answer } from '../api/models'
+import { useApiClients } from '@equinor/fusion'
+
+interface SingleAnswerSummaryProps
+{
+    answer: Answer
+}
+
+const SingleAnswerSummary = ({ answer }: SingleAnswerSummaryProps) => {
+    let apiClients = useApiClients()
```
nicholasdalhaug, line 16:
```suggestion
    const apiClients = useApiClients()
```

frontend/src/views/Evaluation/Alignment/AlignmentView.tsx
```diff
@@ -0,0 +1,76 @@
...
+                    />
+                </Box>
+            </Box>
+
+            { selectedQuestion &&
+                <AnswerSummarySidebar
+                    open={selectedQuestion != undefined}
+                    onCloseClick={() => { setSelectedQuestion(undefined) }}
+                    question={selectedQuestion} />
+            }
```
nicholasdalhaug, line 70: Since the form has an `open` prop, we do not need the nesting. Then we will not get messages about unable to render unmounted component, although they are hard to see among all our other warnings. 

```suggestion
            <AnswerSummarySidebar
                open={selectedQuestion != undefined}
                onCloseClick={() => { setSelectedQuestion(undefined) }}
                question={selectedQuestion} 
            />
```

nicholasdalhaug (changes requested):
I think it would be nice to have a more condensed summary form, like in the figma sketches.

frontend/src/components/BarrierQuestionsView.tsx
tmals-equi: Yeah, I was thinking about this too. But that means deciding if the button is visible based on whether or not that prop is defined. I like that, but I can also see it could be confusing, what do you think?

nicholasdalhaug: Otherwise there will be a coupling between those two props. I think it seems better like you just suggested.

frontend/src/components/AnswerSummarySidebar.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+        <ModalSideSheet
+            header={ barrierToString(question.barrier) }
+            show={open}
+            size='medium'
+            onClose={onCloseClick}
+            isResizable={false}
+        >
+            <div style={{margin: 20}}>
+                <Typography variant="h3">{question.text}</Typography>
```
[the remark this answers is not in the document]
tmals-equi: Yep, I wanted to have it, but it's not trivial to add, because the number seems to be resolved simply by the index during iteration, and that happens in a component further down than this.

frontend/src/components/BarrierQuestionsView.tsx
tmals-equi: :+1:

frontend/src/components/QuestionAndAnswer/QuestionAndAnswerFormWithApi.tsx
```diff
@@ -83,9 +83,11 @@ interface QuestionAndAnswerFormWithApiProps {
     question: Question
     answer: Answer | undefined
     disabled: boolean
+    onQuestionSummarySelected?: (question: Question, questionNumber: number) => void
 }
 
-const QuestionAndAnswerFormWithApi = ({questionNumber, question, answer, disabled}: QuestionAndAnswerFormWithApiProps) => {
+const QuestionAndAnswerFormWithApi = ({questionNumber, question, answer, disabled, onQuestionSummarySelected}: QuestionAndAnswerFormWithApiProps) => {
+    const azureUniqueId = getAzureUniqueId()
```
nicholasdalhaug, line 90: It does not look like this is used.

frontend/src/views/Evaluation/Alignment/AlignmentView.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react'
+import { Barrier, Evaluation, Question, Progression } from '../../../api/models'
+import { Box } from '@material-ui/core'
+import BarrierQuestionsView from '../../../components/BarrierQuestionsView'
+import EvaluationSidebar from '../EvaluationSidebar'
+import { useQuestionsQuery } from '../Preparation/PerparationGQL'
```
nicholasdalhaug, line 6: The code in preparationgql should probably be moved up one level if it is used here as well.

nicholasdalhaug: But I have removed this in my latest code, so that we get the questions from the evaluation query. Maybe we do not need to fix this here, then?

nicholasdalhaug (changes requested):

frontend/src/views/Evaluation/Alignment/AlignmentView.tsx
tmals-equi: OK, if you have removed it then I think it should be fine for now

nicholasdalhaug (approved):

PR #183, opened 2020-12-01
Allow only facilitator to change progression of evaluation

### LINKED ISSUE #156: Backend: Create facilitator attribute

ReedOnly:
closes #156

### DIFF
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+using System;
 using System.Linq;
 
 using api.Services;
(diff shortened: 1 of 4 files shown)


backend/api/GQL/Mutation.cs
```diff
@@ -49,6 +49,11 @@ public Evaluation CreateEvaluation(string name, string projectId)
 
         public Evaluation ProgressEvaluation(string evaluationId)
         {
+            if (!_authService.IsFacilitator(evaluationId))
+            {
+                throw new UnauthorizedAccessException($"Current user is not Facilitator");
+            }
+
             return _evaluationService.ProgressEvaluation(evaluationId);
         }
```
nicholasdalhaug, line 58: Do we not have a test on Progress Evaluation. I think the MockAuthService does not implement `IsFacilitator`. Will that not be a problem?

nicholasdalhaug (approved):
I see a codacy message. It could be nice to have a comment there, but looks good otherwise.

PR #193, opened 2020-12-07
Status tab

tmals-equi:


frontend/src/components/ProgressSummary.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+const getProgressAll = (participant: Participant, questions: Question[]) => {
+    const answeredQuestionsForOrg = questions.filter(q => {
+        return q.answers.filter(a => a.answeredBy?.id === participant.id).length > 0
+    })
+    return Math.ceil((answeredQuestionsForOrg.length / questions.length) * 100)
+}
+
+const ProgressSummary = ({evaluation, participants}: ProgressSummaryProps) => {
+    const {loading: loadingQuestions, questions, error: errorQuestions} = useQuestionsQuery(evaluation.id)
```
nicholasdalhaug, line 93: It should not be necessary to use a new query here. Also, I have removed this whole file in my PR, hope we can get that in first. Do you even need the evaluation? Could you not just take in questions in this component? 

New: `const questions = evaluation.questions`

frontend/src/components/ProgressSummary.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    if(questions === undefined){
+        return <div>
+            <TextArea
+                value={`Questions is undefined`}
+                onChange={() => { }}
+            />
+        </div>
+    }
```
nicholasdalhaug, line 115: All of this can be removed following my comment above. Also, the integration with the cache works better. Of course we will have to refresh the page to get totally up to date answers from the backend for all users.

frontend/src/components/ProgressSummary.tsx
```diff
@@ -0,0 +1,137 @@
+
+import React from 'react'
+import { Evaluation, Organization, Participant, Question } from '../api/models'
+import ParticipantCard from '../views/Evaluation/Nomination/ParticipantCard'
```
nicholasdalhaug, line 4: Are you sure it makes sense to have the `ParticipantCard` in `Nomination` when we use it outside of `Nomination`?

frontend/src/components/StatusTab.tsx
```diff
@@ -0,0 +1,54 @@
...
+import ProgressSummary from './ProgressSummary'
+import { useParticipantsQuery } from '../views/Evaluation/Nomination/NominationGQL'
+
+interface QuestionnaireStatusTabsProps {
+    evaluation: Evaluation
+}
+
+const QuestionnaireStatusTabs = ({evaluation, children}: React.PropsWithChildren<QuestionnaireStatusTabsProps>) => {
+    const [activeTabKey, setActiveTabKey] = React.useState('questionnaire')
+    const {loading: loadingQuery, participants, error: errorQuery} = useParticipantsQuery(evaluation.id)
```
nicholasdalhaug, line 14: The participant is accessible using `useContext`, no need to do a new query. It uses functionality from `views` which does not make sense for something in `components`.

frontend/src/components/ProgressSummary.tsx
```diff
@@ -0,0 +1,137 @@
...
+const getProgressOrganisation = (participant: Participant, questions: Question[]) => {
+    if (participant.organization === Organization.All) {
+        return getProgressAll(participant, questions)
+    }
+
+    const orgQuestions = questions.filter(q => q.organization === participant.organization)
+    const answeredQuestionsForOrg = orgQuestions.filter(q => {
+        return q.answers.filter(a => a.answeredBy === participant).length > 0
+    })
+    return Math.ceil((answeredQuestionsForOrg.length / orgQuestions.length) * 100)
```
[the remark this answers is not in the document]
nicholasdalhaug: My suggestion is also not good enough since it does not filter for the relevant progression. The content of the status page should be different for each step in the stepper.

frontend/src/views/Evaluation/EvaluationView.tsx
```diff
@@ -34,12 +35,12 @@ const EvaluationView = ({evaluation, onProgressEvaluationClick}: EvaluationViewP
                 description={calcProgressionStatus(evaluation.progression, Progression.Preparation)}
                 stepKey={Progression.Preparation}
             >
-                <>
+                <QuestionnaireStatusTabs evaluation={evaluation}>
```
nicholasdalhaug, line 38: If you do this for `Preparation`, should you not also do this for `Alignment`?

nicholasdalhaug (changes requested):

nicholasdalhaug:
One more comment: We should say in this tab wether or not the participant has said they are completed.

nicholasdalhaug (approved):
LGTM 😊

PR #220, opened 2020-12-14
Set up logging, more GQL tests

### LINKED ISSUE #141: Backend: Set up logging

Christdej:
Closes #141

### DIFF
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -11,6 +11,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 using Microsoft.Identity.Web;
 using Microsoft.OpenApi.Models;
 
@@ -28,7 +29,6 @@ namespace api
     public class Startup
     {
         private readonly string _accessControlPolicyName = "AllowSpecificOrigins";
-
         private readonly SqliteConnection _connection;
         private readonly string _sqlConnectionString;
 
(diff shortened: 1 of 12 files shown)


backend/api/Startup.cs
```diff
@@ -145,23 +145,35 @@ public void ConfigureServices(IServiceCollection services)
...
 
+            if (!string.IsNullOrEmpty(_sqlConnectionString))
+            {
+                logger.LogInformation("Using SQL database");
+            }
+            else
+            {
+                logger.LogInformation("Using Sqlite in memory database");
+            }
+
```
nicholasdalhaug, line 164: Shouldn't this logging happen around line 40 instead? That is where we set up the database to be either sql or sqlite.

backend/tests/GQL/Mutation.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+using Xunit;
+
+namespace tests
+{
+    [Collection("UsesDbContext")]
+    public class MutationTest : DbContextTestSetup
+    {
+        private readonly Mutation _mutation;
+        public MutationTest() : base()
```
nicholasdalhaug, line 14: If this is redundant as codacy says, we should remove it.

backend/tests/DbContextTestSetup.cs
```diff
@@ -30,4 +32,14 @@ public void Dispose()
...
         }
     }
+
+    class MockAuthService : IAuthService
+    {
+        public string GetOID()
+        {
+            return "1";
+        }
+        public void AssertIsFacilitator(string evaluationId)
```
nicholasdalhaug, line 42: Maybe do as codacy says here, and explain why it is empty

nicholasdalhaug (commented):
Just some small comments / questions. Looks good otherwise! 👍

backend/api/Startup.cs
Christdej: No, because you cannot inject the logger into Startup or ConfigureServices, it is possible to get a logger there, but it is not recommended if you can avoid it

nicholasdalhaug: 👍

nicholasdalhaug (approved):

PR #249, opened 2021-01-05
Fix env variable issue when deploying to fusion CI

ReedOnly:


nicholasdalhaug:
Should this be a draft PR until fusion has updated its package? Then we can install the new package and this PR can be merged.

ReedOnly:
https://pro-s-portal-ci.azurewebsites.net/apps/bmt

frontend/resources/app-icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nicholasdalhaug, line 1: Could this resources folder be combined with what we have as the docs folder? If it is not simple, we do not need to. But I just thought it could be simpler to get an overview from the root folder then.

pipelines/deploy-fusion-prod.yml
```diff
@@ -0,0 +1,50 @@
+# Deploy fusion-bmt to prod environment
```
nicholasdalhaug, line 1: Will you change the reference to these files as well, so they are used correctly by pipelines and radix? I am thinking about the moving to the `pipelines` folder.

pipelines/deploy-fusion-prod.yml
```diff
@@ -0,0 +1,50 @@
...
+  - deployment: DeployApp
+    environment: fusion
+
+  - job: 'Deploy_Fusion_BMT'
+    steps:
+      - task: Npm@1
+        displayName: 'Build frontend'
+        inputs:
+          command: 'install'
+          workingDir: 'frontend'
```
nicholasdalhaug, line 20:
```suggestion
      - task: Npm@1
        displayName: 'Install frontend'
        inputs:
          command: 'install'
          workingDir: 'frontend'
```

pipelines/deploy-fusion-prod.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - task: FusionApp@2
+        displayName: 'Deploy frontend'
+        inputs:
+          fusionCredentials: 'Fusion BMT'
+          portalUrl: 'https://pro-s-portal-fprd.azurewebsites.net'
+          action: 'Deploy'
+          appKey: 'bmt'
+          bundlePath: 'frontend/out/bmt.zip'
+          tokenResource: '97978493-9777-4d48-b38a-67b0b9cd88d2'
```
nicholasdalhaug, line 37: Is this a secret?

pipelines/deploy-fusion-prod.yml
```diff
@@ -0,0 +1,50 @@
...
+          ignoreVersionConflict: true
+          forceReplaceExisting: true
+
+
+      - task: FusionApp@2
+        displayName: 'Publish frontend'
+        inputs:
+          fusionCredentials: 'Fusion BMT'
+          portalUrl: 'https://pro-s-portal-fprd.azurewebsites.net'
+          action: 'Publish'
```
nicholasdalhaug, line 47: What is the difference between the Deploy and the Publish?

nicholasdalhaug (changes requested):

pipelines/deploy-fusion-prod.yml
ReedOnly: yup, will update after merging 👍

pipelines/deploy-fusion-prod.yml
ReedOnly: nope, is just uuid of ad group

pipelines/deploy-fusion-prod.yml
ReedOnly: you can deploy without actually making visible in the portal. No idea, why its possible btw 😄

nicholasdalhaug (approved):

PR #252, opened 2021-01-06
242: Added filter on org leads for alignment progress

denektenina:


frontend/src/views/Evaluation/EvaluationView.tsx
```diff
@@ -15,6 +15,7 @@ interface EvaluationViewProps {
 }
 
 const EvaluationView = ({evaluation, onProgressEvaluationClick, onProgressParticipant}: EvaluationViewProps) => {
+    let allowedRoles = Object.values(Role);
```
nicholasdalhaug, line 18: As Codacy says here, you should remove the semicolon. 
Also, consider using `const` instead of `let`.

nicholasdalhaug (changes requested):

frontend/src/views/Evaluation/EvaluationView.tsx
denektenina: Omitting semicolons is going to take some time getting used to but I agree 👍 

Will change to const as well, though we might change it when I created it.

nicholasdalhaug (approved):

PR #265, opened 2021-01-08
Added prettier

denektenina:
Needs some discussion about configuration


frontend/.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+    "trailingComma": "all",
+    "tabWidth": 4,
+    "semi": false,
+    "singleQuote": true,
+    "arrowParens": "always"
+}
```
nicholasdalhaug, line 7:
```suggestion
{
    "tabWidth": 4,
    "semi": false,
    "arrowParens": "avoid"
}
```

I suggest we use the `tabWidth` and `semi` as this and avoid the parenthesis around singe argument functions. And we can keep the rest as default. 

[docs](http://json.schemastore.org/prettierrc)

ReedOnly (approved):

ReedOnly (approved):
Looks good!

PR #268, opened 2021-01-11
Frontend: Create View for Edit Action sidebar component

denektenina:


ReedOnly (approved):
Component looks good, and works locally!

frontend/src/components/Action/QuestionActionsList.tsx
```diff
@@ -2,14 +2,21 @@ import React, { useState } from 'react'
...
 import { DataToCreateAction } from '../../api/mutations'
+import styled from 'styled-components'
+
+const ActionLink = styled(Typography)`
+    cursor: pointer;
+
+    &:hover {
+        text-decoration: underline;
+    }
+`
```
nicholasdalhaug, line 19: This would look better if we just used `<Typography link>...`

frontend/src/components/Action/QuestionActionsList.tsx
```diff
@@ -18,58 +25,67 @@ interface Props {
...
-                            Add action
-                        </Button>
+    const editAction = (action: Action) => {
+        setIsSidebarOpen(true)
+        setActionToEdit(action)
+    }
+
+    const onClose = () => {
+        setIsSidebarOpen(false)
+        actionToEdit && setActionToEdit(undefined)
```
nicholasdalhaug, line 39: I cannot remember having seen this used like this before, cool! But I find it more readable and explicit like this: 
```suggestion
        if (actionToEdit !== undefined) {
            setActionToEdit(undefined)
        }
```

I wanted to say that, but if you like that better I might develop a more pronounced preference later.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+import { DataToCreateAction } from '../../api/mutations'
+import styled from 'styled-components'
+import ActionNotesList from './ActionNotesList'
+
+const RightOrientedDate = styled(Typography)`
+    float: right;
+`
+
+const RightOrientedSwitch = styled(Switch)`
+    float: right;
```
nicholasdalhaug, line 20: We have earlier used MUI Box with flex, but this seems better 👍

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+`
+
+type TextFieldChangeEvent = React.ChangeEvent<HTMLTextAreaElement> & React.ChangeEvent<HTMLInputElement>
+
+type Validity = Exclude<TextFieldProps['variant'], undefined | 'warning'>
+
+const ErrorIconComponent = <Icon size={16} data={error_filled} color="danger" />
```
nicholasdalhaug, line 27: I know it might have been there already, but maybe: 

```suggestion
const ErrorIcon = <Icon size={16} data={error_filled} color="danger" />
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+const checkIfTitleValid = (title: string) => {
+    return title.length > 0
+}
+
+const checkIfParticipantValid = (participant: Participant | undefined) => {
+    return participant !== undefined
+}
+
+interface Props {
+    action: Action | undefined
```
nicholasdalhaug, line 38: I know they are equivalent, but: 
```suggestion
    action?: Action
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+        (action && action.assignedTo && action.assignedTo.azureUniqueId) || undefined
+    )
+    const assignedTo: Participant | undefined = possibleAssignees.find(a => a.azureUniqueId === assignedToId)
+    const [assignedToValidity, setAssignedToValidity] = useState<Validity>('default')
+
+    const [dueDate, setDueDate] = useState<Date>((action && action.dueDate) || new Date())
+    const [priority, setPriority] = useState<Priority>((action && action.priority) || Priority.High)
+    const [description, setDescription] = useState<string>((action && action.description) || '')
+    const [onHold, setOnHold] = useState<boolean>(false)
+    const [note, setNote] = useState<string>('')
```
nicholasdalhaug, line 59:
```suggestion
    const [noteInProgress, setNoteInProgress] = useState<string>('')
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+    const assigneesOptions: SearchableDropdownOption[] = possibleAssigneesDetails.map(personDetails => ({
+        title: personDetails.name,
+        key: personDetails.azureUniqueId,
+        isSelected: personDetails.azureUniqueId === assignedToId,
+    }))
+
+    const createdDate = action && action.createDate && new Date(action.createDate).toLocaleString('en-GB').split(',')[0]
+    const isEditMode = action && action.id
```
nicholasdalhaug, line 68:
```suggestion
    const isEditMode = action !== undefined
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const [onHold, setOnHold] = useState<boolean>(false)
+    const [note, setNote] = useState<string>('')
+
+    const assigneesOptions: SearchableDropdownOption[] = possibleAssigneesDetails.map(personDetails => ({
+        title: personDetails.name,
+        key: personDetails.azureUniqueId,
+        isSelected: personDetails.azureUniqueId === assignedToId,
+    }))
+
+    const createdDate = action && action.createDate && new Date(action.createDate).toLocaleString('en-GB').split(',')[0]
```
nicholasdalhaug, line 67:
```suggestion
    const createdDateString = new Date(action?.createDate).toLocaleString('en-GB').split(',')[0]
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+                                onChange={(event: TextFieldChangeEvent) => setNote(event.target.value)}
+                                variant="default"
+                                style={{ height: 75 }}
+                            />
+                        </Grid>
+                        <Grid item xs={1} container={true} alignItems="center">
+                            <Icon data={add} color="color-primary"></Icon>
+                        </Grid>
+                        <Grid item xs={12}>
+                            <ActionNotesList />
```
nicholasdalhaug, line 227: This should be used. And if dummy data is needed, I think it should be placed higher in the hierarchy.

frontend/src/components/Action/ActionNotesList.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Box } from '@material-ui/core'
+
+const notes = [
+    {
+        id: 'hfjkhfkf78f6s9fhj',
+        writerName: 'Anna Therese Asheim',
+        message: 'Kan du se på dette Andreas Rønning?',
+        date: '2021-01-10T11:27:08.024+01:00',
+    },
+]
```
nicholasdalhaug, line 11: I think this should be place higher in the react hierarchy. This component could actually be finished in this PR. 

```suggestion
const notes: Note[] = [
    {
        id: 'hfjkhfkf78f6s9fhj',
        writerName: 'Anna Therese Asheim',
        message: 'Kan du se på dette Andreas Rønning?',
        date: '2021-01-10T11:27:08.024+01:00',
    },
]
```

frontend/src/components/Action/ActionNotesList.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const ActionNotesList = () => {
+    return (
+        <Box mt={2}>
+            {notes.map(note => {
+                const dateAndTime = note && note.date && new Date(note.date).toLocaleString('en-GB').split(',')
+                const date = dateAndTime[0]
+                const hours = dateAndTime[1].split(':')[0]
+                const minutes = dateAndTime[1].split(':')[1]
```
nicholasdalhaug, line 20:
```suggestion
                const noteDate = new Date(note.date)
                
                const date = noteDate.getDate()
                const hour = noteDate.getHours()
                const minutes = noteDate.getMinutes()
```

If we set the types right for the component, we can avoid the `note && note.date &&`.

nicholasdalhaug (changes requested):
You have some nice solutions here that I think we should have used some other places as well, but we can take those as we see fit. I have some comments, and I would like to discuss them as well. 

Not least, could we discuss wether or not to create a totally new sidebar? I think we can avoid some checks at least, and some props. Especially when we start having the possibility to edit actions where there is not possibility to add actions, I think it could be valuable to have a component for just editing an action. Of course, it would give some copy-paste-code, which is really bad. But let us discuss it!

frontend/src/components/Action/QuestionActionsList.tsx
denektenina: Thanks, I agree! I like to use this notation when I check for presence before using an attribute on an object, but in this case, since it's actually checking if it should perform an action, it's better to use the explicit way.

Also, I just found that the way I normally use it (`user && user.name` for instance), can be better written like user?.name, which is pretty cool :) (more info [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining))

frontend/src/components/Action/QuestionActionsList.tsx
denektenina: lol, I didn't know about that. Will fix!

frontend/src/components/Action/ActionForm.tsx
denektenina: In general I prefer being consistent (by using Box since we are using it other places), but in this case it seemed that I had to add additional styling anyways to get the switch all the way to the end, so I ended up with this.

frontend/src/components/Action/ActionForm.tsx
denektenina: Yeah, it was there already but I can change it

frontend/src/components/Action/ActionForm.tsx
denektenina: Yes, that's a lot better. Not quite used to typescript yet! Will fix.

frontend/src/components/Action/ActionForm.tsx
denektenina: Will fix

frontend/src/components/Action/ActionForm.tsx
denektenina: Here I'm actually checking for id because I'm thinking it could maybe exist as an object even though it has not been saved to the database and thus not been given an ID.

frontend/src/components/Action/ActionForm.tsx
denektenina: Why do we have to include the type in the name here?

frontend/src/components/Action/ActionForm.tsx
denektenina: What do you mean?

frontend/src/components/Action/ActionNotesList.tsx
denektenina: I can make the changes but I assume I have to tweak it when connecting to the db since I don't know the real property names.

denektenina: What do you mean by higher in the react hierarchy? In the file tree?

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: That is something that is easier to see if the create and edit action forms were separate. But I think it is safe to say that if the action is not undefined, it will have an id. Otherwise, there is an action in the database with no id, and that is an invalid state. For the action creation, we have no need for this `action` prop.

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Well, we do not have to. But when I read it, I thought it was a date, which it is not, it is a string. `createdDate` sounds like it is a date. On the other hand, the js `Date` is inconsistent on this since if you call `Date.getDate` you get the date number, not all the content of the date object. Could we just do `new Date(action?.createDate).getDate()`?

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug:
```suggestion
                            <ActionNotesList notes={action.notes} />
```

frontend/src/components/Action/ActionNotesList.tsx
nicholasdalhaug: Yes, in the file tree. You can find the real property names in `frontend/src/api/models.ts`.

frontend/src/components/Action/ActionForm.tsx
denektenina: Yeah it's a habit from coding forms earlier. Often you would for instance make an empty object and work with that before saving, and then the only thing that separates a saved and new object is whether or not it has been sent to the db and thus been given an Id. But since we are using state for the attributes before saving here we can use your way. Anyways, I guess the new form-or not thing is something for the later discussion? I don't mind changing it if we decide it's the best :)

frontend/src/components/Action/ActionForm.tsx
denektenina: I split the string to omit the time and only keep the date. It is possible to get each of the components of the date (like month, day and year) from the Date object of course, but since the GB-way of formatting was so close to what we needed I thought it was the best way of doing it. I can include string in the name if you feel it makes it easier to understand the code.

frontend/src/components/Action/ActionForm.tsx
denektenina: Yes, I will fix rest of the notes-things when I implement the backend functionality. This was just for the view.

frontend/src/components/Action/ActionNotesList.tsx
denektenina: I started on this but it requires some more work than just selecting two different property names (because of the Notes type). It's not a big job, but I might as well do it when I start digging into the Notes-universe in the next task later today.

frontend/src/components/Action/ActionNotesList.tsx
denektenina: If it's ok with you I will also fix this as part of the  other task

nicholasdalhaug (approved):

PR #270, opened 2021-01-11
Test startup.cs and swagger

### LINKED ISSUE #271: Component test

Christdej:
Closes #271

### DIFF
--- a/backend/api/appsettings.json
+++ b/backend/api/appsettings.json
@@ -1,25 +1,25 @@
 {
-  "AzureAd": {
-    "Instance": "https://login.microsoftonline.com",
-    "Authority": "https://sts.windows.net/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/",
-    "Audience": "8829d4ca-93e8-499a-8ce1-bc0ef4840176",
-    "Domain": "equinor.com",
-    "TenantId": "3aa4a235-b6e2-48d5-9195-7fcf05b459b0",
-    "ClientId": "8829d4ca-93e8-499a-8ce1-bc0ef4840176",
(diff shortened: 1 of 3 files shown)


backend/api/appsettings.json
```diff
@@ -1,25 +1,25 @@
...
-  },
-  "ApplicationInsights": {
-    "InstrumentationKey": ""
-  },
-  "Logging": {
-    "LogLevel": {
-      "Default": "Information",
-      "Microsoft": "Warning",
-      "Microsoft.Hosting.Lifetime": "Information"
+    "AzureAd": {
```
ReedOnly, line 2: whats wrong with this file?

backend/tests/StartupTest.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class ComponentTest
+    {
+        private readonly WebApplicationFactory<Startup> factory = new WebApplicationFactory<Startup>();
+
+        [Fact]
+        public async Task StartupTest()
+        {
+            var client = factory.CreateClient();
+
+            var response = await client.GetAsync("/swagger/index.html");
```
ReedOnly, line 19: This only checks if swagger exists, not if API works right?

ReedOnly (approved):

backend/api/appsettings.json
Christdej: Changing from 2 spaces to 4, because I like 4 spaces better

backend/tests/StartupTest.cs
Christdej: This checks startup.cs, and that swagger returns 200-299 code. If you mess something up in startup.cs, this test will fail

PR #273, opened 2021-01-11
Create and use SeveritySummary

nicholasdalhaug:


frontend/src/components/SeverityIndicator.tsx
```diff
@@ -34,4 +62,40 @@ const SeverityIndicator = ({severity}: SeverityIndicatorProps) => {
...
+interface SeverityIndicatorWithNumberProps {
+    severity: Severity
+    n: number
+}
+
+export const SeverityIndicatorWithNumber = ({n, severity}: SeverityIndicatorWithNumberProps) => {
+    return <>
+        <div style={{
+            display: "grid",
+            placeItems: "center",
```
denektenina, line 74: Burde vi hatt en bestemmelse om hvorvidt vi skal bruke inline styling eller styled components når vi styler på denne måten?

frontend/src/components/SeveritySummary.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Box } from "@material-ui/core"
+
+import { Severity } from "../api/models"
+import SeverityIndicator, { getBGColor, getTextColor, SeverityIndicatorWithNumber } from "./SeverityIndicator"
+import { Typography } from "@equinor/eds-core-react"
+
+
+
+interface SeveritySummaryFieldProps {
+    n: number
```
denektenina, line 12: Burde vi bruke variabelnavn som ikke bare er én bokstav? (mindre lesbart som det er nå)

denektenina (approved):

frontend/src/components/SeverityIndicator.tsx
nicholasdalhaug: What do you think? I think they can both have their advantages and disadvantages, so maybe we do not want to use only one of them. But if you have an other view on it?

frontend/src/components/SeveritySummary.tsx
nicholasdalhaug: I guess we could use `num`, but `number` is reserved. Suggestions?

frontend/src/components/SeverityIndicator.tsx
denektenina: I tend to prefer styled components, especially because it lets us give the components names that describe their function / style more precisely (like ClickableAction), but I don't like having many lines of styled components at the top of the file. But either way we don't seem to use too much styling since we get a lot from EDS and material UI, so I think having the little styling we do as styled components is good.

frontend/src/components/SeveritySummary.tsx
denektenina: Even though its also an abbreviation I prefer num over n if number is reserved. It could also be useful to think about what the number represents and give it a name from its function rather than what it is. Such as severityIndicator instead of number.

ReedOnly (approved):

frontend/src/components/SeverityIndicator.tsx
denektenina: I like being able to style the components "inline" using props in material UI, but I thought maybe it was considered more difficult to maintain to style inline with CSS?

nicholasdalhaug: That is a good point! Do you think we should stop using MUI? It might be prettier using inline styles for styling that is dynamic, right? Anyways, I think such a discussion can be done elsewhere, maybe in an issue of it self?

frontend/src/components/SeveritySummary.tsx
nicholasdalhaug: Yes, will rename to `num`. I was trying to think of what it was, but for the perspective of the component, I think it is actually just a number. I was considering `count`, but I think it is too specific for that component.

frontend/src/components/SeverityIndicator.tsx
denektenina: I think Material UI gives us a lot of styling for free that saves us from writing that css, so in my opinion we should keep it as long as we are using it, and it works for our purpose. The way I understand it, using styling through props (such as `<Box display="flex" flexDirection="column">`) is OK because we use style that is separated from our code, but using inline style (such as `<div style={display: flex}>`) is not as good, because we risk not being able to find the styling that easy when we need to change it, and it overrides other styles (for instance if we define a common style for all divs in a css-file). Not sure if it would be a problem for us, and also as long as the styles aren't many I think it's actually kind of handy to have it "right there", but yeah. Maybe it should be a bigger discussion. It wasn't that easy to find resources discussing this either. A lot of the pages saying inline css was bad was from 2010, so maybe it's not that relevant anymore.

frontend/src/components/SeveritySummary.tsx
denektenina: Ok, thanks for clarifying!

PR #294, opened 2021-01-14
Integrated "edit action" and "notes" view with backend

denektenina:


frontend/package-lock.json
```diff
@@ -4948,6 +4948,16 @@
...
+        "bindings": {
+            "version": "1.5.0",
+            "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
+            "integrity": "sha512-p2q/t/mhvuOj/UeLlV6566GD/guowlr0hHxClI0W9m7MWYkL1F0hLo+0Aexs9HSPCtR1SXQ0TD3MMKrXZajbiQ==",
+            "dev": true,
+            "optional": true,
+            "requires": {
+                "file-uri-to-path": "1.0.0"
+            }
+        },
```
nicholasdalhaug, line 4960: Why did the package-lock change? Just wondering since we do not want to change it back and forth.

frontend/src/api/fragments.ts
```diff
@@ -79,6 +79,21 @@ export const ACTION_FIELDS_FRAGMENT = gql`
...
+    fragment NoteFields on Note {
+        id
+        text
+        createDate
+        createdBy {
+            azureUniqueId
+        }
+        action {
+            ...ActionFields
+        }
```
nicholasdalhaug, line 92: I do not think we should have the whole action available on all notes. Consider just adding it to the mutation later.

frontend/src/api/mutations.ts
```diff
@@ -26,6 +31,13 @@ interface EditActionMutationProps {
     error: ApolloError | undefined
 }
 
+interface CreateNoteMutationProps {
+    createNote: (data: DataToCreateNote) => void
```
nicholasdalhaug, line 35: I understand we did this in the same way as above, but we have mostly been using simple arguments. Then we do not have to create objects and pass them in, only fill out the arguments. Like this: 
`createNote: (actionId: string, text: string) => void`

frontend/src/api/mutations.ts
```diff
@@ -130,9 +142,52 @@ export const useEditActionMutation = (): EditActionMutationProps => {
...
         action: data?.editAction,
         error,
     }
 }
+
+export const useCreateNoteMutation = (): CreateNoteMutationProps => {
+    const CREATE_NOTE = gql`
+        mutation CreateNote($text: String, $actionId: String) {
+            createNote(text: $text, actionId: $actionId) {
+                ...NoteFields
```
nicholasdalhaug, line 156: Since the mutation wants the action id of the note, it makes sense to ask for that here: 
`action { id }`

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -20,6 +20,10 @@ const RightOrientedSwitch = styled(Switch)`
     float: right;
 `
 
+const ClickableIcon = styled(Icon)`
+    cursor: pointer;
+`
```
nicholasdalhaug, line 27: If the Icon should be clickable, could we not just make it a button with an Icon in? I believe we have done that other places.

frontend/src/components/Action/QuestionActionsList.tsx
```diff
@@ -7,29 +7,30 @@ import { add } from '@equinor/eds-icons'
...
-        setActionToEdit(action)
+        setActionToEditId(action.id)
     }
 
     const onClose = () => {
         setIsSidebarOpen(false)
-        if (actionToEdit) {
-            setActionToEdit(undefined)
+        if (actionToEditId) {
+            setActionToEditId(undefined)
```
nicholasdalhaug, line 33: Would it make more sense to just set it to undefined anyways? Or maybe it makes it less understandable? I just think it would be nice to not have to understand the if-statement.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -82,6 +85,19 @@ const ActionForm = ({
...
     const isEditMode = action !== undefined
 
+    useEffect(() => {
+        const timeout = setTimeout(() => {
+            autosaveChanges()
+        }, WRITE_DELAY_MS)
+        return () => {
+            clearTimeout(timeout)
+        }
+    }, [title, description])
```
nicholasdalhaug: Consider making this into our own hook so it is easier to reuse. I believe this is reused in part from Q&A form, but also saves on mount.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -82,6 +85,19 @@ const ActionForm = ({
...
+
+    useEffect(() => {
+        autosaveChanges()
+    }, [assignedTo, onHold, dueDate, priority, completed])
```
nicholasdalhaug: Consider not saving on mount

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -168,10 +186,17 @@ const ActionForm = ({
...
-                        <Grid item xs={8}>
-                            <Button onClick={onCompleteEditClick}>Complete</Button>
+                        <Grid item xs={6}>
+                            <Switch
+                                checked={completed}
+                                onChange={() => {}} // This is required to avoid an error
+                                onClick={() => setCompleted(!completed)}
+                                disabled={false}
+                                label="Completed"
+                                enterKeyHint=""
```
nicholasdalhaug, line 198: I guess the comment should be here as well: `// This is required to avoid an error`

nicholasdalhaug (commented):
For some reason, all my comments on earlier commits have disappeared. I will come back to this PR later.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -85,18 +85,20 @@ const ActionForm = ({
...
 
+    const firstUpdate = useRef(true)
     useEffect(() => {
+        if (firstUpdate.current === true) {
+            firstUpdate.current = false
+            return
+        }
+
         const timeout = setTimeout(() => {
             autosaveChanges()
```
nicholasdalhaug, line 96: See Codacy

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -294,6 +296,7 @@ const ActionForm = ({
                                 onChange={(event: TextFieldChangeEvent) => setNoteInProgress(event.target.value)}
                                 onKeyPress={(e: any) => {
                                     if (e.key === 'Enter') {
+                                        e.preventDefault()
```
nicholasdalhaug, line 299: Should this be added in the earlier commit?

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -85,18 +85,20 @@ const ActionForm = ({
     const createdDateString = new Date(action?.createDate).toLocaleDateString()
     const isEditMode = action !== undefined
 
+    const firstUpdate = useRef(true)
     useEffect(() => {
+        if (firstUpdate.current === true) {
+            firstUpdate.current = false
+            return
+        }
+
```
nicholasdalhaug: Should this be added in the earlier commit?

frontend/src/components/Action/ActionNotesList.tsx
```diff
@@ -9,24 +9,40 @@ interface Props {
 }
 
 const ActionNotesList = ({ notes, participantsDetails }: Props) => {
+    const sorted = notes.slice().sort((a, b) => {
```
nicholasdalhaug, line 12: `sortedNotes` maybe. A bit more readable I think.

nicholasdalhaug (changes requested):
I realised in the middle of this that this PR should probably be squashed. 

I ran the code and the page looks really good!

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -114,13 +130,30 @@ const ActionForm = ({ action, connectedQuestion, possibleAssignees, possibleAssi
         }
     }
 
+    const onCompleteEditClick = () => {
+        const valid = setFormValidity()
+        if (valid && action) {
+            const actionData: Action = {
```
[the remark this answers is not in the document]
denektenina: But it's not a new action, right? It's just the data for updating the existing action?

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -114,13 +130,30 @@ const ActionForm = ({ action, connectedQuestion, possibleAssignees, possibleAssi
...
+                ...action,
+                title,
+                onHold,
+                dueDate,
+                priority,
+                description,
+                completed: false, // TODO: This is still to be decided
```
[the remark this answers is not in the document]
denektenina: This is now fixed (I'm now updating with autosave, so clicking complete (which is now a switch and not a button), will set completed to true or false, depending on the previous value)

frontend/package-lock.json
denektenina: True. I think it happened because I ran npm install after having pulled, or maybe when I installed the project on my new mac. Don't know why this was updated for me first though. I have not installed anything new. Was unsure if I should commit this file. Do you want me to remove it from the PR?

frontend/src/api/fragments.ts
denektenina: Will just add action id then.

frontend/src/components/Action/ActionForm.tsx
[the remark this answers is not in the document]
nicholasdalhaug: ´editedAction`?

frontend/package-lock.json
nicholasdalhaug: I just checked and I get the same here. Then I think it will not change all the time and we should keep it.

frontend/src/api/mutations.ts
denektenina: Will fix

frontend/src/api/mutations.ts
denektenina: Will fix

frontend/src/components/Action/ActionForm.tsx
denektenina: Good idea! Will fix

frontend/src/components/Action/QuestionActionsList.tsx
denektenina: I guess it wouldn't make a practical difference, so I can do that

frontend/src/components/Action/ActionForm.tsx
denektenina: As discussed, we will fix this when your code is ready

frontend/src/components/Action/ActionForm.tsx
denektenina: Have fixed

frontend/src/components/Action/ActionForm.tsx
denektenina: Fixed

frontend/src/components/Action/ActionNotesList.tsx
denektenina: Fixed

frontend/src/components/Action/ActionForm.tsx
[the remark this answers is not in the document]
denektenina: Ok will fix this and the create Action function above (with newAction)

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Is it not a problem since it is only run after mount?

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Are you sure? 
```suggestion
                                enterKeyHint="" // This is required to avoid an error
```

nicholasdalhaug (approved):

PR #312, opened 2021-01-19
Save indicator in edit action-view

denektenina:


frontend/src/components/Action/ActionForm.tsx
```diff
@@ -13,6 +13,24 @@ import styled from 'styled-components'
...
+`
+
+const SmallerCircularProgress = styled(CircularProgress)`
+    width: 25px;
+    height: 25px;
+`
+
+const SaveText = styled(Typography)`
+    padding-left: 10px;
+`
```
nicholasdalhaug, line 33: I still think this could be placed as its own component. 

The `SmallCircularProgress` I can understand, even though I am not sure if it is necessary to move the styling to the top of the file. But the other components  here I do not see how that the names imply the styling, what type of components they are based on the name nor why they are placed at the top of the file. The code would be easier to read if the padding for example is simply placed in the Typography where it is used. 

Also, is it necessary to specify top and right here? That sounds like styling based on the use of the component and not styling for the component itself. 

Should we use `rem` instead of `px`?

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -36,8 +54,15 @@ const checkIfParticipantValid = (participant: Participant | undefined) => {
     return participant !== undefined
 }
 
+const SavingState = {
+    SAVING: 'SAVING',
+    SAVED: 'SAVED',
+    NONE: 'NONE',
+}
+
```
nicholasdalhaug, line 62: This could be a typescript enum. And I think we should name them with PascalCasing, eg. `Saving`.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -36,8 +54,15 @@ const checkIfParticipantValid = (participant: Participant | undefined) => {
...
 interface Props {
     action?: Action
+    actionIsSaving: boolean
```
nicholasdalhaug, line 65: I would call it `isActionSaving`, but do you think this is better? I find it easier to know exactly what it is when it starts with "is". Then I immediately see it is a boolean and can read the rest of the name to see what it represents. `actionIsSaving` sounds like a statement to me and I have to think a bit. But I think it sounds better in an if-statement: `if (actionIsSaving) {`.

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Maybe call the enum `SavingStates` since it is not a state but rather all the possible states? Or maybe it is good when `savingState` is an implementation of `SavingState`, not sure.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -139,6 +166,21 @@ const ActionForm = ({
...
+                setSavingState(SavingState.SAVED)
+                setTimeout(() => {
+                    setSavingState(SavingState.NONE)
+                }, SAVETEXT_FADEOUT_DELAY)
+            } else {
+                setSavingState(SavingState.NONE)
+            }
+        }
+    }, [actionIsSaving])
+
```
nicholasdalhaug, line 183: If you time this perfectly, I think you can get a situation where it says "Saved" right after you write and while it is saving. This is because the timeout does not check if it is still not saving after 1 second. Right? I think it can be fixed with a cleanup function in the useEffect.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -212,7 +267,10 @@ const ActionForm = ({
                         value={title}
                         autoFocus={true}
                         label="Title"
-                        onChange={(event: TextFieldChangeEvent) => setTitle(event.target.value)}
+                        onChange={(event: TextFieldChangeEvent) => {
+                            setSavingState(SavingState.SAVING)
```
nicholasdalhaug, line 271: We do not need to set saving state on different fields. It is set by the input variable `isActionSaving`.

frontend/src/components/Action/ActionForm.tsx
denektenina: True. I suggest we remove it. It's not a big problem, it seems to end up with the correct confirmation after having flashed through an extra status, but since it's unpredictable and not completely necessary, I suggest we just keep the saved-icon to confirm to the user that the action is saved after changes.

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -45,6 +46,7 @@ interface Props {
     onActionEdit: (action: Action) => void
     onNoteCreate: (actionId: string, text: string) => void
     onCancelClick: () => void
+    setSavingState: (savingState: number) => void
```
nicholasdalhaug, line 49: I do not think the form need to know about the saving states. Could we just have a function `onErrorValidity` or something?

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -149,6 +152,7 @@ const ActionForm = ({
             if (!isParticipantValid) {
                 setAssignedToValidity('error')
             }
+            setSavingState(SavingState.None)
```
nicholasdalhaug, line 155:
```suggestion
            onErrorValidity()
```

frontend/src/components/Action/ActionForm.tsx
```diff
@@ -212,7 +216,10 @@ const ActionForm = ({
                         value={title}
                         autoFocus={true}
                         label="Title"
-                        onChange={(event: TextFieldChangeEvent) => setTitle(event.target.value)}
+                        onChange={(event: TextFieldChangeEvent) => {
+                            setSavingState(SavingState.Saving)
+                            setTitle(event.target.value)
+                        }}
```
nicholasdalhaug, line 222:
```suggestion
                        onChange={(event: TextFieldChangeEvent) => {
                            setTitle(event.target.value)
                        }}
```

nicholasdalhaug: When saving can be figured out on the outside of this component, maybe in the `onActionEdit` function.

frontend/src/components/SaveIndicator.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect } from 'react'
+import { CircularProgress, Icon, Typography } from '@equinor/eds-core-react'
+import { check_circle_outlined } from '@equinor/eds-icons'
+import { SavingState } from '../utils/Variables'
+import styled from 'styled-components'
+
+interface SaveIndicatorProps {
+    isSaving: boolean
+    savingState: number
+    setSavingState: (savingState: number) => void
```
nicholasdalhaug, line 10:
```suggestion
    savingState: SavingState
    setSavingState: (savingState: SavingState) => void
```

frontend/src/components/SaveIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+        if (isSaving) {
+            setSavingState(SavingState.Saving)
+        } else {
+            if (savingState === SavingState.Saving) {
+                setSavingState(SavingState.Saved)
+            } else {
+                setSavingState(SavingState.None)
+            }
+        }
+    }, [isSaving])
```
nicholasdalhaug, line 29: It seems for me like there is a spread of responsibilities here. I think this component does not need to change the saving state. It takes in a saving state and displays the correct content. Deciding what that saving state is could be done on the outside, instead of both outside and inside this component.

nicholasdalhaug (changes requested):

frontend/src/components/SaveIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+        <>
+            {savingState === SavingState.Saving && (
+                <SaveConfirmation>
+                    <CircularProgress style={{ width: '25px', height: '25px' }}></CircularProgress>
```
nicholasdalhaug, line 35:
```suggestion
                    <CircularProgress style={{ width: '25px', height: '25px' }} />
```

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Agreed

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Sorry, this is to make it say "Saving" before the actual call to the api, during the 1s delay.

frontend/src/components/Action/ActionForm.tsx
denektenina: Yes, that's probably better!

frontend/src/components/Action/ActionForm.tsx
denektenina: This is fixed by making new component and moving some of the styling to the tags. Have not changed to rem, but is it any point since we are using px other places?

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: I think this is difficult to fix without splitting edit and create form

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Well, we still need the setSavingState to run to get the "Saving" before the edit call is done. So I think this should not be done. 
I think this is difficult to fix without splitting edit and create form

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: I think this is difficult to fix without splitting edit and create form

frontend/src/components/Action/ActionForm.tsx
nicholasdalhaug: Yes, I think we might want to change it the other placed as well 😊

denektenina: Ok. Then resolve maybe?

frontend/src/components/SaveIndicator.tsx
denektenina: Ok, yeah, I agree that it's not ideal to do it both places. Only thing is that this code is a bit ugly and would be good to hide one place instead of repeating throughout the application

nicholasdalhaug (approved):

PR #326, opened 2021-01-21
Add health endpoint to check if backend and db is running

### LINKED ISSUE #325: Backend: Create Health endpoint for backend

ReedOnly:
closes #325

### DIFF
--- a/backend/api/.editorconfig
+++ b/backend/api/.editorconfig
@@ -11,3 +11,6 @@ dotnet_diagnostic.IDE0044.severity = none
 
 # IDE0039: Use local function instead of lambda
 dotnet_diagnostic.IDE0039.severity = none
+
+# ASP0001: Call UseAuthorization before app.UseEndpoints
+dotnet_diagnostic.ASP0001.severity = none
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -88,12 +88,12 @@ namespace api
             if (!string.IsNullOrEmpty(_sqlConnectionString))
             {
                 // Setting splitting behavior explicitly to avoid warning
-                services.AddDbContext<BmtDbContext>(options => options.UseSqlServer(_sqlConnectionString, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SingleQuery)), ServiceLifetime.Transient);
+                services.AddDbContext<BmtDbContext>(options => options.UseSqlServer(_sqlConnectionString, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SingleQuery)));
             }
             else
             {
                 // Setting splitting behavior explicitly to avoid warning
-                services.AddDbContext<BmtDbContext>(options => options.UseSqlite(_connection, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SingleQuery)), ServiceLifetime.Transient);
+                services.AddDbContext<BmtDbContext>(options => options.UseSqlite(_connection, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SingleQuery)));
             }
 
             services.AddErrorFilter<ErrorFilter>();
@@ -121,7 +121,6 @@ namespace api
 
             services.AddApplicationInsightsTelemetry();
             services.AddHealthChecks().AddCheck<EvaluationService>("ModelsFromDB");
-            services.AddHealthChecks().AddDbContextCheck<BmtDbContext>();
 
             services.AddSwaggerGen(c =>
             {
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -119,6 +119,9 @@ namespace api
 
             services.AddControllers();
 
+            services.AddApplicationInsightsTelemetry();
+            services.AddHealthChecks().AddDbContextCheck<BmtDbContext>();
+
             services.AddSwaggerGen(c =>
             {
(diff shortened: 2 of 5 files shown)


backend/api/Startup.cs
```diff
@@ -119,6 +119,9 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddControllers();
 
+            services.AddApplicationInsightsTelemetry();
```
Christdej, line 122: ApplicationInsight is added in Program.cs, is this needed?

Christdej (changes requested):

nicholasdalhaug (commented):
Have you checked if this gives an error for the case of unfit database to the models used?

backend/api/Startup.cs
ReedOnly: This will give more metrics as well, not only the logs

ReedOnly:
It gives error when database is down, but not if it is missing a migration. The problem is also that not all migrations make changes that will be affected by a particular test querying the db.

backend/api/Startup.cs
Christdej: As long as it does not add noise to the logs Im happy

nicholasdalhaug (approved):

backend/api/Startup.cs
```diff
@@ -88,12 +88,12 @@ public void ConfigureServices(IServiceCollection services)
             if (!string.IsNullOrEmpty(_sqlConnectionString))
             {
                 // Setting splitting behavior explicitly to avoid warning
-                services.AddDbContext<BmtDbContext>(options => options.UseSqlServer(_sqlConnectionString, o => o.UseQuerySplittingBehavior(QuerySplittingBehavior.SingleQuery)));
```
ReedOnly, line 91: I simplified this and added Transient lifetime for db connection, in order to do polling for health check. Is there any reason we needed the querySPlittingBehaviour?

Christdej: This was a bug that was introduced in HotChocolate v5, It look like they have fixed in 5.0.2, will dobbel check

backend/api/Startup.cs
```diff
@@ -119,6 +119,10 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddControllers();
 
+            services.AddApplicationInsightsTelemetry();
+            services.AddHealthChecks().AddCheck<GraphQuery>("GraphQl");
```
Christdej, line 123: Rename the healthcheck to "ModelsFromDB" or something I think, because this is not actually checking GraphQL, but rather that we can retrieve values from the DB and turn them into models

backend/api/GQL/Query.cs
```diff
@@ -102,5 +109,20 @@ public IQueryable<Note> GetNotes()
         {
             return _noteService.GetAll();
         }
+
+        public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
```
Christdej, line 113: I think the interface allows us to not take HealthCheckContext as a parameter, so make Codacy happy

backend/api/GQL/Query.cs
```diff
@@ -102,5 +109,20 @@ public IQueryable<Note> GetNotes()
...
+        {
+            try
+            {
+                Evaluation evaluation = _evaluationService.GetAll().First();
```
Christdej, line 117: Check if it is possible to only do:
_evaluationService.GetAll().First();
(remove the left side of the equal sign)
to see if this still fails on the test db, no need to assign to a variable if we dont use it

Christdej (changes requested):

backend/api/.editorconfig
```diff
@@ -11,3 +11,6 @@ dotnet_diagnostic.IDE0044.severity = none
 
 # IDE0039: Use local function instead of lambda
 dotnet_diagnostic.IDE0039.severity = none
+
+# ASP0001: Call UseAuthorization before app.UseEndpoints
+dotnet_diagnostic.ASP0001.severity = none
```
nicholasdalhaug, line 16: Why do we not want to have this warning? Why not just move the `useAuthorization`?

backend/api/Startup.cs
nicholasdalhaug: Is the querysplitting behaviour fixed? And what does transient lifetime mean? I thought the connection had to be up at all times for sqlite?

backend/api/Startup.cs
```diff
@@ -191,6 +195,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
                 c.OAuthAdditionalQueryStringParams(new Dictionary<string, string>
                     { { "resource", $"{Configuration["AzureAd:ClientId"]}" } });
             });
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapHealthChecks("/health");
+            });
```
nicholasdalhaug, line 201: Are we creating endpoints without authorization? Is that safe?

backend/api/.editorconfig
ReedOnly: because we dont want to have auth on the /health endpoint

backend/api/Startup.cs
ReedOnly: No, still not fixed

nicholasdalhaug (approved):

PR #338, opened 2021-01-22
Fix 'Saving' showing in create action

nicholasdalhaug:


ReedOnly (approved):

denektenina (approved):
Looks good! Also tested it locally. I prefer the name `autosave` to `save`, because then the developer gets additional information about how it works without having to read everything, but that's a personal preference so you don't have to change it back if you disagree.

nicholasdalhaug:
I think the name should reflect more what it does and not how it is used. That the useEffect effectively is the autosave that uses the save. I will merge this.

PR #344, opened 2021-01-25
Saved-icon always present on existing actions

denektenina:


frontend/src/components/Action/ActionSidebar.tsx
```diff
@@ -46,11 +46,13 @@ const ActionSidebar = ({
         } else {
             if (savingState === SavingState.Saving) {
                 setSavingState(SavingState.Saved)
+            } else if (actionExists) {
+                setSavingState(SavingState.Saved)
             } else {
                 setSavingState(SavingState.None)
```
nicholasdalhaug, line 52: Now the else will never be called, right? Also, react does not like when a variable is not placed in the dependency array. Maybe we do not need this functionality now?

frontend/src/components/Action/ActionSidebar.tsx
```diff
@@ -46,11 +46,13 @@ const ActionSidebar = ({
...
             }
         }
-    }, [isActionSaving])
+    }, [isActionSaving, open])
```
nicholasdalhaug, line 55: I do not think we need to run the code every time the sidesheet opens or closes.

nicholasdalhaug (changes requested):

frontend/src/components/Action/ActionSidebar.tsx
denektenina: Unfortunately we need to run it the first time, because the icon is set to None before the panel opens, so if we want it to show, and to only show for existing actions, this does the trick. It doesn't open and close that often, so I don't think it's a problem running the code that sets state whenever the panel opens (or closes)

frontend/src/components/Action/ActionSidebar.tsx
denektenina: I think it's good to have an else to show the "default" value in case neither of the conditions apply? And I don't understand what you mean by your last comment, can you explain?

denektenina:
Closing this PR because it does not have an issue and we didn't see an immediate need for the functionality

PR #360, opened 2021-01-28
Order question

### LINKED ISSUE #108: Backend: Make QuestionTemplates editable
This includes, but not exclusively, 
* Testing service.Create
* Testing that ActiveQuestions does not return inactive questions

### LINKED ISSUE #286: Backend: Store questions with weighting based on excel order

Christdej:
Closes #286 
Closes #108


backend/api/Services/QuestionTemplateService.cs
```diff
@@ -54,5 +55,70 @@ string supportNotes
...
+
+        public QuestionTemplate ReorderQuestionTemplate(QuestionTemplate questionTemplate, QuestionTemplate newNextQuestionTemplate = null)
+        {
+            List<QuestionTemplate> questionTemplates = _context.QuestionTemplates.OrderBy(qt => qt.Order).ToList();
+
+            int newOrder = newNextQuestionTemplate.Order - 1;
+            if (newNextQuestionTemplate.Equals(null))
+            {
+                newOrder = _context.QuestionTemplates.Max(qt => qt.Order);
+            }
```
nicholasdalhaug, line 98: I am guessing this would give an error on the line above if the if is true. 

```suggestion
            int newOrder;
            if (newNextQuestionTemplate.Equals(null))
            {
                newOrder = _context.QuestionTemplates.Max(qt => qt.Order);
            }
            else 
            {
                newOrder = newNextQuestionTemplate.Order - 1;
            }
```

backend/tests/Services/QuestionTemplateService.cs
```diff
@@ -33,5 +34,71 @@ public void ActiveQuestions()
...
+        public void ReorderQuestionTemplate()
+        {
+            QuestionTemplateService questionTemplateService = new QuestionTemplateService(_context);
+            IQueryable<QuestionTemplate> getAll = questionTemplateService.GetAll();
+            List<int> initialOrder = getAll.Select(qt => qt.Order).ToList();
+            HashSet<int> initialOrderSet = new HashSet<int>(initialOrder);
+            Assert.Equal(initialOrder.Count, initialOrderSet.Count);
+
+            QuestionTemplate questionTemplate = getAll.ToList()[3];
+            int prevOrder = questionTemplate.Order;
```
nicholasdalhaug, line 96: Maybe use `prevOrder`?

nicholasdalhaug (changes requested):
Looks good! I just had one thing that looked like a potential error. Maybe we should have a test for that small edge case.

Christdej (commented):
Make migration and add to PR

ReedOnly:
The migration we remove, should weremove it from git is this pr as well, or wait until after db update is completed?

backend/api/Services/QuestionTemplateService.cs
Christdej: yes, fixed

backend/tests/Services/QuestionTemplateService.cs
Christdej: removed

ReedOnly (approved):

ReedOnly (approved):

PR #362, opened 2021-01-29
Suggested refactor: Made a new QuestionList component to share code for listing questions

denektenina:


frontend/src/components/QuestionsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+import QuestionActionsListWithApi from './Action/QuestionActionsListWithApi'
+import AnswerSummaryButton from './AnswerSummaryButton'
+import { useParticipant } from '../globals/contexts'
+
+type Props = {
+    questions: Question[]
+    viewProgression: Progression
+    disable: boolean
+    displayActions?: boolean
+    useFacilitatorAnswer?: boolean
```
nicholasdalhaug, line 15: In this context I think it is more accurate to say

```suggestion
    useAnyAnswer?: boolean
```

frontend/src/components/QuestionsList.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const QuestionsList = ({
+    questions,
+    viewProgression,
+    disable,
+    onQuestionSummarySelected,
+    displayActions = false,
+    useFacilitatorAnswer = false,
+}: Props) => {
+    const { azureUniqueId: participantUniqueId } = useParticipant()
```
nicholasdalhaug, line 27: It is misleading to say that it is the id of the participant. The participant id is the one in our database, not the one that is globally unique to that user. 

```suggestion
    const { azureUniqueId: currentUserAzureUniqueId } = useParticipant()
```

frontend/src/components/QuestionsList.tsx
denektenina: Yes, that was my first naming, but after you explained I felt this name gave more meaning because it explains the reasoning behind the logic (i.e. we are not using users answers on this step, but the answers/summaries from the facilitators).

frontend/src/components/QuestionsList.tsx
denektenina: I don't think this was my naming, but I can change it.

frontend/src/components/QuestionsList.tsx
nicholasdalhaug: Ok, but then maybe `useOnlyFacilitatorAnswers´?

nicholasdalhaug (approved):

PR #364, opened 2021-01-29
Made view for actions list

denektenina:


frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
+import React, { useEffect, useState } from 'react'
```
nicholasdalhaug, line 1: Shouldn't this file be in "components". It will be reused later, so I think it should be reusable.

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+            }
+        })
+    })
+
+    return { azureUniqueIds, actions }
+}
+
+interface Props {
+    evaluation: Evaluation
+}
```
nicholasdalhaug, line 50: I am pretty sure that the action list only should take in actions and list them. I do not see why the actions list itself would extract actions from an evaluation.

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    const getAllPersonDetails = (azureUniqueIds: string[]): Promise<PersonDetails[]> => {
+        const manyPromises: Promise<PersonDetails>[] = azureUniqueIds.map(azureUniqueId => {
+            return apiClients.people.getPersonDetailsAsync(azureUniqueId).then(response => {
+                return response.data
+            })
+        })
+
+        return Promise.all(manyPromises)
+    }
```
nicholasdalhaug, line 67: Isn't this just a copy of other code we have?

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+            setPersonDetailsList(fetchedPersonDetailsList)
+        })
+    }, [questions])
+
+    const assignedPersonDetails = (action: Action) => {
+        const assignedToId = action.assignedTo?.azureUniqueId
+        return personDetailsList.find((p: PersonDetails) => p.azureUniqueId === assignedToId)
+    }
+
+    useEffect(() => {
```
nicholasdalhaug, line 84: Consider using `useMemo` instead.

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+                        }
+                        case 'priority':
+                            return sortPriority(a[accessor], b[accessor], sortDirection)
+                        default:
+                            return sort(a[accessor], b[accessor], sortDirection)
+                    }
+                })
+            )
+        }
+    }, [sortDirection, columnToSortBy])
```
nicholasdalhaug, line 105: Not `actions`?

frontend/src/utils/sort.ts
```diff
@@ -0,0 +1,33 @@
+import { Priority } from '../api/models'
+
+export type SortDirection = 'ascending' | 'descending' | 'none'
+
+export const sort = (a: string, b: string, sortDirection: SortDirection) => {
```
nicholasdalhaug, line 5:
```suggestion
export const sortString = (a: string, b: string, sortDirection: SortDirection) => {
```

Or include number, boolean, etc. as input types.

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { sort, SortDirection, sortPriority } from '../../../utils/sort'
+
+const { Body, Row, Cell, Head } = Table
+
+const PriorityDisplay = styled.div`
+    display: flex;
+    flex-direction: row;
+`
+
+type ColumnType = {
```
nicholasdalhaug, line 18:
```suggestion
type Column = {
```

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+        })
+    }, [questions])
+
+    const assignedPersonDetails = (action: Action) => {
+        const assignedToId = action.assignedTo?.azureUniqueId
+        return personDetailsList.find((p: PersonDetails) => p.azureUniqueId === assignedToId)
+    }
+
+    useEffect(() => {
+        if (columnToSortBy) {
```
nicholasdalhaug, line 85: What does this mean? Can it be undefined?

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+                    switch (accessor) {
+                        case 'assignedTo': {
+                            const assignedToA = assignedPersonDetails(a)
+                            const assignedToB = assignedPersonDetails(b)
+
+                            return assignedToA && assignedToB ? sort(assignedToA.name, assignedToB.name, sortDirection) : 0
+                        }
+                        case 'priority':
+                            return sortPriority(a[accessor], b[accessor], sortDirection)
```
nicholasdalhaug, line 98: You do not need to use `accessor` when you know it is `priority`.

frontend/src/views/Project/ProjectRoute.tsx
```diff
@@ -64,9 +63,6 @@ const ProjectRoute = ({ match }: RouteComponentProps<Params>) => {
                 <Tab tabKey="dashboard" title="Dashboard">
                     <ProjectDashboardView project={project} />
                 </Tab>
-                <Tab tabKey="actions" title="Actions">
-                    <ProjectActionsView />
-                </Tab>
```
nicholasdalhaug, line 69: Maybe remove archive as well?

nicholasdalhaug (changes requested):
Remove empty file "ActionSidebar.tsx"

I see that the commits should be squashed. What will be the commit message?

frontend/src/utils/sort.ts
denektenina: good point, will put number and boolean as input types

frontend/src/utils/sort.ts
```diff
@@ -0,0 +1,33 @@
...
+    } else if (
+        (a === Priority.High && b === Priority.Low) ||
+        (a === Priority.Medium && b === Priority.Low) ||
+        (a === Priority.High && b === Priority.Medium)
+    ) {
+        return sortDirection === 'ascending' ? 1 : -1
+    }
+
+    return 0
+}
```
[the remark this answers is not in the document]
denektenina: fixed

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
denektenina: Fixed by making a hook that all three components are now using

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
denektenina: done

frontend/src/views/Evaluation/FollowUp/ActionList.tsx
denektenina: fixed by using useMemo and including action there

frontend/src/views/Evaluation/FollowUp/SortableHeadCell.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
```
nicholasdalhaug, line 1: Consider moving into ActionTable component

nicholasdalhaug (changes requested):

nicholasdalhaug (approved):

PR #368, opened 2021-02-05
Fix action assigned to and created by mixup

Christdej:


backend/api/Services/ActionService.cs
```diff
@@ -17,8 +17,8 @@ public ActionService(BmtDbContext context)
         }
 
         public Action Create(
-            Participant assignedTo,
             Participant createdBy,
+            Participant assignedTo,
             string description,
             DateTimeOffset dueDate,
             string title,
```
Christdej, line 24: From Mutation.cs the create action service is called as following:

`return _actionService.Create(CurrentUser(evaluation), assignedTo, description, dueDate, title, priority, question);`
Instaed of changing the order the parameters are sent in, we change to order in the service to a more "logical" order

nicholasdalhaug (approved):

ReedOnly (approved):

PR #383, opened 2021-02-11
Ordered questions by new order-attribute on question

denektenina:


frontend/src/views/Evaluation/Individual/IndividualView.tsx
```diff
@@ -90,7 +91,7 @@ const IndividualView = ({ evaluation, onNextStepClick, onProgressParticipant }:
                         </Box>
                     </Box>
                     <QuestionsList
```
nicholasdalhaug, line 93: Can we simply order them right before they are being used instead, so in the `QuestionsList`?

nicholasdalhaug (changes requested):

frontend/src/views/Evaluation/Individual/IndividualView.tsx
denektenina: Yes, that's better!

nicholasdalhaug (approved):

PR #406, opened 2021-04-21
Follow up summary view

denektenina:


frontend/src/views/Evaluation/FollowUp/Summaries/Bowtie.tsx
```diff
@@ -110,23 +110,23 @@ const Text = styled(Typography)`
 `
 
 interface Props {
-    answersWithBarrier: AnswersWithBarrier[]
+    answersWithBarrier: BarriersWithAnswers[]
```
maiaht, line 113: Maybe the prop name should be in the same order as the name of the interface, such that `answersWithBarrier` -> `barrierWithAnswers`?

frontend/src/views/Evaluation/FollowUp/Summaries/BowtieView.tsx
```diff
@@ -17,18 +17,18 @@ const ScrollableDiv = styled.div`
...
+const BowtieView = ({ barriersWithProgressionAnswer }: Props) => {
     return (
         <>
             <Box p="20px">
                 <Typography variant="h2" style={{ marginBottom: '30px' }}>
                     Bowtie model
                 </Typography>
                 <ScrollableDiv>
-                    <Bowtie answersWithBarrier={answersWithBarrier} />
+                    <Bowtie answersWithBarrier={barriersWithProgressionAnswer} />
```
maiaht, line 31: `answersWithBarrier` -> `barrierWithAnswers` here as well?

frontend/src/views/Evaluation/FollowUp/Summaries/Bowtie.tsx
denektenina: Yes, that actually was the old name. Which is probably why it is still the name in the interface (?) the reason I changed it is that when it's created we go through a list of barriers and assign answers to it, so it felt more correct to highlight the barrier-part of it, but I'm not very attached to this change so I can change it back

frontend/src/views/Evaluation/FollowUp/Summaries/BowtieView.tsx
denektenina: I can change the order, but I want to keep the word progression in there to show better what it is

maiaht (approved):

PR #452, opened 2021-06-15
Workflow summary for facilitators

### LINKED ISSUE #439: #15 Summary field for workshop
A text field where the X-role can conclude/summaries the workshop session

Q: Which role should be able to write the summary?

ErlendHaa:
A couple of points worth nothing:

 - Currently _only_ the facilitators can modify the summary
 - The summary field is a string on the ``Evaluation``. I.e. we have no timestamps or record of authors. 
 
Closes #439

### DIFF
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -77,6 +77,15 @@ namespace api.GQL
             return evaluation;
         }
 
+        public Evaluation SetSummary(string evaluationId, string summary)
+        {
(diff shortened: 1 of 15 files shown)


maiaht (commented):
Have you set up Prettier for your editor?

ReedOnly:
This should only be visible for Facilitator and Admin

ReedOnly (changes requested):
Looks good, but after talking to users, they want that this summary should only be visible for Facilitator and Admin

ReedOnly (approved):

PR #453, opened 2021-06-15
Let facilitators share answers in workshop/followup

ErlendHaa:
Important note: The behaviour when 2 or more facilitators are editing the same answer _at the same time_ is undefined. There is also no warning issued to the users about this


ErlendHaa:
Closed #454

ReedOnly (approved):
Remove export default for useSharedFacilitatorAnswer

ReedOnly (approved):

PR #458, opened 2021-06-23
Link previous evaluation

### LINKED ISSUE #446: #146 Link to previous evaluation
It should be possible to create a link to a previous evaluation within a project, when an evaluation is created.

For now the link is simply stored in our database, and can be used in Power BI reports.

ErlendHaa:
Give users the option to add a link to a previous evaluation when
creating a new evaluation. For now this link is not used anywhere else
in the frontend.

Run `dotnet ef database update` on dev database after merging

Closes #446

### DIFF
--- a/frontend/cypress/integration/create_evaluation_spec.ts
+++ b/frontend/cypress/integration/create_evaluation_spec.ts
@@ -1,0 +1,110 @@
+import { Role } from '../../src/api/models'
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { evaluationName } from '../support/helpers'
+import NominationPage from '../support/nomination'
+import ProjectPage from '../support/project'
+import users, { User } from '../support/users'
+
+
+const evaluationUserIsFacilitator = () => {
+    let seed = new EvaluationSeed({
+        nParticipants: 3,
+        namePrefix: 'user is Facilitator'
+    })
+    let user  = seed.participants[2]
+    user.role = Role.Facilitator
+    return seed
+}
+
+const evaluationUserIsParticipant = () => {
+    return new EvaluationSeed({
+        nParticipants: 3,
+        namePrefix: 'user is Participant'
+    })
+}
+
+const evaluationUserIsNotInIt = () => {
+    return new EvaluationSeed({
+        nParticipants: 2,
+        namePrefix: 'user is not part of this Evaluation'
+    })
+}
+
+
+describe('Creating a new Evaluation', () => {
+    const user: User = users[2]
+    let userIsFacilitator: EvaluationSeed
+    let userIsParticipant: EvaluationSeed
+    let userIsNotInEvaluation: EvaluationSeed
+
+    before( () => {
+        cy.login()
+        cy.interceptExternal()
+        userIsFacilitator = evaluationUserIsFacilitator()
+        userIsParticipant = evaluationUserIsParticipant()
+        userIsNotInEvaluation = evaluationUserIsNotInIt()
+
+        userIsFacilitator.plant()
+        userIsParticipant.plant()
+        userIsNotInEvaluation.plant()
+    })
+
+    beforeEach( () => {
+        cy.login(user)
+        const port = Cypress.env('FRONTEND_PORT') || '3000'
+        cy.visit(`http://localhost:${port}/123`)
+        cy.wait(1000) //const wait is not good, but don't know how to make it stable
+    })
+
+    it('Without setting a previous evaluation', () => {
+        const name = evaluationName({prefix: 'CreatedWithoutPrevLink'})
+
+        const projectPage = new ProjectPage()
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.createEvaluation(name)
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
+
+    it('Choosing a previous evaluation where user is Participant', () => {
+        const name = evaluationName({prefix: 'CreatedWithPrevLink'})
+
+        const projectPage = new ProjectPage()
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.createEvaluation(name, userIsParticipant.name)
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
+
+    it('Choosing a previous evaluation where user is Facilitator', () => {
+        const name = evaluationName({prefix: 'CreatedWithPrevLink'})
+
+        const projectPage = new ProjectPage()
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.createEvaluation(name, userIsFacilitator.name)
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
+
+    it('Choosing a previous evaluation that user is not part of', () => {
+        const name = evaluationName({prefix: 'CreatedWithPrevLink'})
+
+        const projectPage = new ProjectPage()
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.createEvaluation(name, userIsNotInEvaluation.name)
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
+})
--- a/frontend/cypress/support/project.ts
+++ b/frontend/cypress/support/project.ts
@@ -17,5 +17,19 @@ export default class ProjectPage {
              */
             return cy.get('[data-testid=create_evaluation_dialog_create_button_grid]').then($el => cy.wrap($el).find('button'))
         }
+
+        previousEvaluation = () => {
+            return cy.contains('Previous Evaluation').parent()
+        }
+
+        createEvaluation = (name: string, previousEvaluation?: string) => {
+            this.nameTextField().type(name)
+
+            if (previousEvaluation) {
+                this.previousEvaluation().click().type(`${previousEvaluation}{enter}`)
+            }
+
+            this.createButton().click()
+        }
     }
 }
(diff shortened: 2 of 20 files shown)


ReedOnly (approved):
**Questions:**
* How will already created evaluations behave, will they have undefined previousEvaluation?
* For a first evaluation there is no previous evaluation, can you set this field to empty when creating that evaluation?

ErlendHaa:
Q:

When choosing a previous evaluation from the dropdown list, which evaluations (from current project) should be shown:
1. all Evaluations
2. all Evaluations which the current users is part of 
3. all Evaluations that is created by the current user

ErlendHaa:
Require:  #488

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+            const name = evaluationName({prefix: 'CreatedWithPrevLink'})
+
+            const projectPage = new ProjectPage()
+            projectPage.createEvaluationButton().click()
+
+            const dialog = new ProjectPage.CreateEvaluationDialog()
+            dialog.nameTextField().type(name)
+            dialog.previousEvaluation().click().type(`${userIsNotInEvaluation.name}{enter}`)
+            dialog.createButton().click()
+        })
```
achaikou, line 110: Make it a method? I had hard time spotting the difference

frontend/cypress/support/project.ts
```diff
@@ -17,5 +17,9 @@ export default class ProjectPage {
              */
             return cy.get('[data-testid=create_evaluation_dialog_create_button_grid]').then($el => cy.wrap($el).find('button'))
         }
+
+        previousEvaluation= () => {
+            return cy.get('.fc--TextInput__inputTextContent--1gUuS')
```
achaikou, line 22: Is it possible to use a better selector?

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
+import { Role } from '../../src/api/models'
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { evaluationName } from '../support/helpers'
+import NominationPage from '../support/nomination'
+import ProjectPage from '../support/project'
+import users, { User } from '../support/users'
+
```
achaikou, line 7: I thought you wanted to go with spec convention?
evaluations_spec.ts?

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+const evaluationUserIsParticipant = () => {
+    return new EvaluationSeed({
+        nParticipants: 3,
+        namePrefix: 'user is Participant'
+    })
+}
+
+const evalutionUserIsNotInIt = () => {
```
achaikou, line 26: evalu(a)tion

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+            const projectPage = new ProjectPage()
+            projectPage.createEvaluationButton().click()
+
+            const dialog = new ProjectPage.CreateEvaluationDialog()
+            dialog.nameTextField().type(name)
+            dialog.previousEvaluation().click().type(`${userIsFacilitator.name}{enter}`)
+            dialog.createButton().click()
+        })
+
+        it('Evaluation where user is not a participanting is selectable', () => {
```
achaikou, line 100: participa(nt)

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+const evalutionUserIsNotInIt = () => {
+    return new EvaluationSeed({
+        nParticipants: 2,
+        namePrefix: 'user is not part of this Evaluation'
+    })
+}
+
+
+describe('Creating new Evaluations', () => {
+    const user: User = users[2]
```
achaikou, line 35: Note for future: 
We should do a better design with regards to chosen users. It's was too difficult now to see who we log in with and why tests work the way they work.
Unless someone does it before me, I hope to try and deal with it in the near future.

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+        it('Evaluation where user is not a participanting is selectable', () => {
+            const name = evaluationName({prefix: 'CreatedWithPrevLink'})
+
+            const projectPage = new ProjectPage()
+            projectPage.createEvaluationButton().click()
+
+            const dialog = new ProjectPage.CreateEvaluationDialog()
+            dialog.nameTextField().type(name)
+            dialog.previousEvaluation().click().type(`${userIsNotInEvaluation.name}{enter}`)
+            dialog.createButton().click()
```
achaikou, line 109: Same to all: are we checking "selectable" of "creation"

If creation, wait until we navigate to the next page?  We want to check that creation happens, right? Even if user is not a participant of previous evaluation.

If selectable, do not create?

frontend/cypress/integration/creating_evaluations.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+describe('Creating new Evaluations', () => {
+    const user: User = users[2]
+    let userIsFacilitator: EvaluationSeed
+    let userIsParticipant: EvaluationSeed
+    let userIsNotInEvaluation: EvaluationSeed
+
+    before( () => {
```
achaikou, line 40: That runs twice for some reason, right after runner is started.
On fresh environment (first test) there are two evaluations of each kind.
It's possible to see it through debugger.

Don't know what causes it, but we may deal with it later.

achaikou (commented):
Took a look at Cypress stuff only

frontend/cypress/integration/creating_evaluations.ts
ErlendHaa: Agree! But not gonna do it here

frontend/cypress/integration/creating_evaluations.ts
ErlendHaa: Not on my watch?

frontend/cypress/integration/create_evaluation_spec.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        const projectPage = new ProjectPage()
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.createEvaluation(name, userIsNotInEvaluation.name)
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
```
achaikou, line 110: OK :smile:  By "make it a method, it's hard to spot the difference", I meant to move all this code to method, like
```
it('Choosing an evaluation where user is participant ')
{
   the_full_test_steps({previous: userIsParticipant}
}
it('Choosing an evaluation user is not part of ')
{
   the_full_test_steps({previous: userIsNotInEvaluation}
}
```

Other option is just to repeat the step several times. The one below works, but I think it should be possible somehow to avoid additional setup? I looks like code in "context" is processed before code in "before", so in simplified version objects are undefined. But maybe you with your experience can see that way to make it simpler, as in concept I like that more than previous suggestion?

```
    class Thing {
        seed: EvaluationSeed | undefined
        name: string

        constructor({name} : {name: string}) {
            this.name = name
        }
    }

    const user: User = users[2]
    let userIsFacilitator = new Thing({name: 'user is Facilitator'})
    let userIsParticipant = new Thing({name: 'user is Participant'})
    let userIsNotInEvaluation = new Thing({name: 'user is not in evaluation'})
    let previousEvaluations = [userIsFacilitator, userIsParticipant, userIsNotInEvaluation]

    before(() => {
        ...
        userIsFacilitator.seed = evaluationUserIsFacilitator()
        userIsFacilitator.seed.plant()
        ...
    })

    context('Can create evaluation ', () => {
        previousEvaluations.forEach(previous => {
            it(`Choosing a previous evaluation where ${previous.name}`, () => {
                ...
                dialog.createEvaluation(name, previous.seed!.name)
               ...
            })
        })
    })

```

frontend/cypress/support/project.ts
```diff
@@ -21,5 +21,14 @@ export default class ProjectPage {
         previousEvaluation= () => {
```
achaikou, line 21: Format this, btw

frontend/cypress/integration/create_evaluation_spec.ts
```diff
@@ -0,0 +1,111 @@
...
+        projectPage.createEvaluationButton().click()
+
+        const dialog = new ProjectPage.CreateEvaluationDialog()
+        dialog.nameTextField().type(name)
+        dialog.createButton().click()
+
+        const nominationPage = new NominationPage()
+        nominationPage.evaluationTitle().should('have.text', name)
+    })
+
```
achaikou, line 72: Is there btw a way to check that previous evaluation was added? I am not sure what's its purpose, so wondering

frontend/cypress/support/project.ts
```diff
@@ -21,5 +21,14 @@ export default class ProjectPage {
...
             return cy.get('.fc--TextInput__inputTextContent--1gUuS')
```
achaikou, line 22: If we want to get rid of it, options:

 ```
<Grid item xs={12} data-testid="create_evaluation_dialog_previous_evaluation_grid">
<SearchableDropdown
label="Previous Evaluation (Optional)"
```
1.
```
        previousEvaluation = () => {
            return cy.contains('Previous Evaluation').parent()
        }
```
```
                this.previousEvaluation().click().type(`${previousEvaluation}{enter}`)
```

2. 

```
        previousEvaluation2 = () => {
            return cy.get('[data-testid=create_evaluation_dialog_previous_evaluation_grid]').then($el => cy.wrap($el).find('input'))
        }
```
```
                this.previousEvaluation2().parent().click().type(`${previousEvaluation}{enter}`)
```

3.
```
        typePreviousEvaluation = (previousEvaluation: string) => {
            return cy.get('[data-testid=create_evaluation_dialog_previous_evaluation_grid]').within(() => {
                cy.get('input').parent().click().type(`${previousEvaluation}{enter}`)
            })
        }
```
```
this.typePreviousEvaluation(`${previousEvaluation}{enter}`)
```

All use some component internal logic, so not good. Not sure if much better than choosing by style. 
First seems simplest though.

achaikou (commented):
Now I also start getting these weird sql errors.
hm. Shall see if that's the problem in real tests. Probably yes, as we shall be using in memory db there too.
not good

PR #467, opened 2021-07-07
Enable facilitator to edit individual assessment after workshop has started

### LINKED ISSUE #463: Facilitator should be able to edit individual assessment after workshop has started

hnformentin:
Closes #463 

The commit statement seems slightly long, happy to change if needed.
@ReedOnly it would be informative if we could review it together, in case it is not the expected behaviour I can get direct feedback. Thanks.

### DIFF
--- a/frontend/src/views/Evaluation/Individual/IndividualView.tsx
+++ b/frontend/src/views/Evaluation/Individual/IndividualView.tsx
@@ -101,7 +101,7 @@ const IndividualView = ({ evaluation, onNextStepClick, onProgressParticipant }: 
                         questions={barrierQuestions}
                         organizationFilter={organizationFilter}
                         viewProgression={viewProgression}
-                        disable={disableAllUserInput || isParticipantCompleted}
+                        disable={(disableAllUserInput || isParticipantCompleted) && participantRole !== Role.Facilitator}
                     />
                 </Box>
             </Box>


frontend/src/views/Evaluation/Individual/IndividualView.tsx
```diff
@@ -101,7 +101,7 @@ const IndividualView = ({ evaluation, onNextStepClick, onProgressParticipant }:
                         questions={barrierQuestions}
                         organizationFilter={organizationFilter}
                         viewProgression={viewProgression}
-                        disable={disableAllUserInput || isParticipantCompleted}
+                        disable={(disableAllUserInput && !Role.Facilitator) || (isParticipantCompleted && !Role.Facilitator)}
```
ErlendHaa, line 104: Shouldn't this be: 
`{(disableAllUserInput || isParticipantCompleted) && participantRole !== Role.Facilitator }` 
?

hnformentin: Thanks @ErlendHaa 
Both seems to result in the same change, but I agree your suggestion is more synthetic. 
If `Role.Facilitator` is a boolean, why do we need to compare the `participantRole !== Role.Facilitator`?
E.g. wouldn't this be even better?
`{(disableAllUserInput || isParticipantCompleted) && !Role.Facilitator }`

ReedOnly (approved):

PR #489, opened 2021-07-20
Delete action

### LINKED ISSUE #443: #20 We should be able to delete an action
Q: Which users/roles should be able to delete

achaikou:
closes #443 

It still fails randomly asserting that deleted action is found in DOM, even though it is clearly deleted. Cypress waits for state to change, so reason for the failure is unclear.
Might be that it will behave more stable on CI, then locally, but shall see.

Upd: regarding Codacy.
Looks like .eslintrc.json is not picked up by codacy at all while it works locally with eslint.

It might be that Codacy just doesn't pick the files up. We might have to explicitly define instead all the rules we need (and Codacy lacks export rules feature, so it might be painful as we don't exactly know what settings were used), put them into eslint file in root directory, merge, and then request Codacy to take rules from config file itself. In that case they will have to be common for all the directories.
If that's so, Codacy settings will be taken as a separate PR.

Our other options are:
- explicit comments (it works, by would be tiring) 
- disable the no-undef check completely in UI (which should be fast and easy). After all, we do typescript which should pick up undeclared variables on its own


backend/api/GQL/Mutation.cs
```diff
@@ -185,6 +185,16 @@ public Action EditAction(string actionId, string assignedToId, string descriptio
             return _actionService.EditAction(action, assignedTo, description, dueDate, title, onHold, completed, priority);
         }
 
+        public Action DeleteAction(string actionId)
+        {
+            /* Note that no related fields are loaded */
+            IQueryable<Action> queryableAction = _actionService.GetAction(actionId);
+            Action action = queryableAction.First();
```
ErlendHaa, line 192: IQueriable.First throws InvalidOperationException if sequence is empty [1]

[1] https://docs.microsoft.com/en-us/dotnet/api/system.linq.queryable.first?view=net-5.0

frontend/src/api/mutations.ts
```diff
@@ -150,6 +150,59 @@ export const useEditActionMutation = (): EditActionMutationProps => {
...
+        update(cache, mutationResult) {
+            const deleteAction = mutationResult.data.deleteAction
+            cache.modify({
+                id: cache.identify({
+                    __typename: 'Question',
+                    id: questionId,
+                }),
+                fields: {
+                    actions(existingActionRefs, { readField }) {
+                        return existingActionRefs.filter((actionRef: any) => {
```
ErlendHaa, line 185: I guess the apollo cache does not cascade deletions? We probably have dangling notes in the cache. Probably not worth fixing?

frontend/src/components/Action/QuestionActionsList.tsx
```diff
@@ -98,6 +119,19 @@ const QuestionActionsList = ({ question, participants }: Props) => {
                 connectedQuestion={question}
                 possibleAssignees={participants}
             />
+            <ConfirmationDialog
+                isOpen={isConfirmDeleteDialogOpen}
+                title="Delete Action?"
+                description="Deleting an action will remove it and its notes from Workshop and Follow-up tabs."
```
ErlendHaa, line 126: `... will permanently delete it ... from the evaluation.` ?

frontend/cypress.json
```diff
@@ -1 +1,3 @@
-{}
+{
+    "chromeWebSecurity": false
```
ErlendHaa, line 2: Why?

frontend/cypress/integration/actions_spec.ts
```diff
@@ -0,0 +1,97 @@
...
+            seed.addAction(actionToDelete).addAction(actionToStay)
+            return { seed, actionToDelete, actionToStay }
+        }
+
+        let seed: EvaluationSeed
+        let actionToDelete: Action
+        let actionToStay: Action
+
+        beforeEach(() => {
+            ;({ seed, actionToDelete, actionToStay } = createSeed())
```
ErlendHaa, line 37: `const { seed, actionToDelete, actionToStay } = createSeed()` ?

frontend/cypress/integration/actions_spec.ts
```diff
@@ -0,0 +1,97 @@
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { Priority, Progression } from '../../src/api/models'
+import { Action } from '../support/mocks'
+import ActionsGrid from '../support/action_grid'
+import { ConfirmationDialog } from '../support/common'
+import { DELETE_ACTION } from '../support/gql'
+
+describe('Actions', () => {
+    context('Delete', () => {
+        const createSeed = () => {
```
ErlendHaa, line 11: Should we decide on where to define such helpers? I would opt for putting everything that doesn't _need_ to be inside a describe, outside of it. Then the actual test cases pop a bit more

backend/api/GQL/Mutation.cs
achaikou: Yep, but code around is also using First, not FirstOrDefault?

Do you have a better suggestion what to use here? If element is not available we want to fail? So why not to use the default error?

frontend/cypress.json
achaikou: We get an "Cypress detected a cross origin error happened" error when reloading the page. I assume fusion tries to do something under the hood
https://docs.cypress.io/guides/guides/web-security#Limitations

frontend/cypress/integration/actions_spec.ts
achaikou: Agreed to wait with precise agreement, but will move createSeed below all the tests, as contents of databse bar returned values are not important

frontend/src/api/mutations.ts
achaikou: We have Notes and deleted Action still in cache, and they are not removed by cache.gc().
No idea why, but at least both objects are not visible. Maybe I misunderstand, and Action and Note are actually also considered root objects, and must be removed with evict?
No idea how they ended up there as well.

Maybe that's something we'll be able to solve with more caching experience :woman_shrugging:

ReedOnly:
Regarding .eslintrc.json, we have to manually enable it in Codacy settings to read that file, but then it will _only_ follow the rules from the config file, and it looks like the file need to be in repository root as well. I suggest that we disable no undef check for frontend, because like you said Typescript will inform about that anyways.

frontend/cypress/support/mocks.ts
```diff
@@ -61,47 +51,64 @@ export class Answer {
...
+        createdBy = faker.random.arrayElement(seed!.participants),
+        // no access to questions as creation is run before plant. Reconsider?
+        questionOrder = faker.datatype.number({ min: 1, max: 3 }),
+        createDate = faker.date.past(),
+        dueDate = faker.date.future(),
+        title = faker.lorem.sentence(),
+        priority = faker.random.arrayElement(Object.values(Priority)),
+        description = faker.lorem.words(),
+        completed = faker.datatype.boolean(),
+        onHold = faker.datatype.boolean(),
```
ErlendHaa, line 97: I would argue it's a bit weird to pass in seed here with the sole purpose of setting default values. Would suggest giving this responsibility to the createAction function, and leave the constructor as it was

ErlendHaa (approved):

PR #491, opened 2021-07-21
Update npm dependencies

### LINKED ISSUE #473: BMT: Update frontend dependencies to solve security vulnerabilities

ReedOnly:
closes #473


ReedOnly:
Removed redundant dependency react-dom and fusion-component

ErlendHaa (approved):

frontend/package.json
```diff
@@ -27,7 +27,6 @@
         "@equinor/eds-core-react": "^0.8.4",
         "@equinor/eds-icons": "^0.6.2",
         "@equinor/eds-tokens": "^0.5.4",
-        "@equinor/fusion-components": "^2.1.0",
```
achaikou, line 30: Why did you decide to remove this dependency as direct dependency?
We explicitly use it when importing.

Is it an ok practice?? I can't find anything that supports it or is against it.

achaikou (commented):
As you explicitly asked for my review and then decided to merge while I was in the middle or figuring out what happens, deal with it now! :smile:

README.md
```diff
@@ -10,7 +10,7 @@ To run the project with docker-compose use:
 ### Prerequisites
 
 -   [.NET 5.0+](https://dotnet.microsoft.com/download/dotnet/5.0)
--   [Node 12+ with npm](https://github.com/nodesource/distributions/blob/master/README.md)
+-   [Node 14+ with npm](https://github.com/nodesource/distributions/blob/master/README.md)
```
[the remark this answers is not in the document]
ReedOnly: We've been using lockfile version 2 for some time, it is just the readme that was not updated, And it was released from npm v7 which came from node 15, which means the readme is still not up to date!

frontend/package.json
ReedOnly: Because we got a get a warning from webpack that we include multiple versions of fusion-components, and fusion-cli already has a version of fusion-components. I see your point that it might be confusing to use a package not specified in package.json. But we also risk using a version of fusion-componets out of sync with fusion-cli, so I think it's better to stick with the one provided there 😃

PR #501, opened 2021-07-29
Add Cypress to CI

### LINKED ISSUE #466: BMT: Set up CI for cypress E2E tests

ReedOnly:
closes #466
- Add Cypress E2E on PR builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,19 @@ cd backend/api
 dotnet run
 ```
 
+## Testing
+We are using Cypress as a test framework for End to End tests. Details can be found
+in [this section](frontend/cypress/README.md).
+
+Cypress E2E tests can be run locally with:
+`docker-compose -f docker-compose.cypress.yml up cypress`
+
+To run locally provide CYPRESS_RECORD_KEY in docker compose file.
+
+Cypress tests will be run in Azure DevOps when pushing to the upstream branch cypress.
+This can be done with the following command:
+`git push upstream HEAD:cypress -f`
+
 ## Configuration
 
 ### Database configuration
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Build Status](https://dev.azure.com/Shellvis/Fusion-BMT/_apis/build/status/equinor.fusion-bmt?branchName=master)](https://dev.azure.com/Shellvis/Fusion-BMT/_build/latest?definitionId=22&branchName=master)
 [![Codacy Badge](https://app.codacy.com/project/badge/Grade/45edba07b87447489c54a51867141261)](https://www.codacy.com/gh/equinor/fusion-bmt/dashboard?utm_source=github.com&utm_medium=referral&utm_content=equinor/fusion-bmt&utm_campaign=Badge_Grade)
+[![BMT](https://img.shields.io/endpoint?url=https://dashboard.cypress.io/badge/detailed/ry3x7y&style=flat&logo=cypress)](https://dashboard.cypress.io/projects/ry3x7y/runs)
 
 To run the project with docker-compose use:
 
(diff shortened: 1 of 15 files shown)


frontend/cypress.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM cypress/base:10
```
achaikou, line 1: I recall you said something that cypress takes long to install (?). Not sure how or why, but if that's the problem then maybe we can speed up using the cypress//included image?

pipelines/azure-pipelines.yml
```diff
@@ -1,62 +1,64 @@
 trigger:
-- master
+    - master
+
+pr:
+    - master
```
achaikou, line 5: Don't get this change.

You explicitly want to run checks only when PR is done to master, right, not to whatever branch? Do we every create PRs to whatever branches?

achaikou (commented):
Looked only at the code today, will try to run locally and check real runs tomorrow, but now looks great!

pipelines/azure-pipelines.yml
ReedOnly: this says that we should run pr builds to master explicitly. We did it before as well, but then it was not specified in the pipeline file.

frontend/cypress.Dockerfile
ReedOnly: Yes, but the included image is crazy big, so it turned out to be even slower!

README.md
```diff
@@ -73,6 +73,19 @@ cd backend/api
...
 ```
 
+## Testing
+We are using Cypress as a test framework for End to End tests. Details can be found
+in [this section](frontend/cypress/README.md).
+
+Cypress E2E tests can be run locally with:
+`docker-compose -f docker-compose.cypress.yml up cypress`
+
+To run locally provide CYPRESS_RECORD_KEY in docker compose file.
```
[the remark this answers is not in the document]
ReedOnly: Its the cypress key for recording. But for local runs we don't really need it, so a better option is to remove line 14 and 15 from cypress.Dockerfile

README.md
```diff
@@ -73,6 +73,20 @@ cd backend/api
 dotnet run
 ```
 
+## Testing
+We are using Cypress as a test framework for End to End tests. Details can be found
+in [this section](frontend/cypress/README.md).
+
+Cypress E2E tests can be run locally with:
+`docker-compose -f docker-compose.cypress.yml up cypress`
```
achaikou, line 81: 1. why `... up cypress` ? Doesn't it mean that you start only cypress server? 
2. Does that just work for you when run from the root directory? I get
`Could not find a MSBuild project file in '/source/api'.` from backend and `Missing script: "start"` from frontend

frontend/cypress/plugins/index.js
```diff
@@ -18,9 +18,8 @@
 // eslint-disable-next-line no-unused-vars
 module.exports = (on, config) => {
     /* Update if we are not running on localhost in CI */
```
achaikou, line 20: Remove the comment?

achaikou (commented):
Codacy makes sense, I still can't run it locally, but consider approved beside this

frontend/cypress/plugins/index.js
ReedOnly: fixed!

README.md
ReedOnly: 1. Since it is specified in docker-compose file that cypress depends on frontend, and frontend depends on backend etc it will start all services, but only cypress will write to console
2. It should work to run from everywhere, I just ran it from backend folder as well

achaikou (approved):
Please don't forget to change the name of the frontend_URL, but otherwise looks great!! :tada:

Issue #474, opened 2021-07-07
BMT: Save date for when workshop has been finished in database


ReedOnly:
This is needed for #138

hnformentin:
Isn't it need for #430 instead?
That would be great to have some additional description, if you have something else in mind let me know.
My guess so far is that it should be in the backend folder, so I am investigating :-)

hnformentin:
Understanding the reference PR #452 
Backend:
- backend/api/GQL/Mutation.cs: what is the convention to use "_"?
- backend/api/Migrations/20210709134613_WorkshopSummary.Designer.cs and
 backend/api/Migrations/20210709134613_WorkshopSummary.cs: autogenerated files that enable compatibility between evaluations created earlier to the changes
- backend/api/Migrations/BmtDbContextModelSnapshot.cs: autogenerated file (1) when should product version be changed? (2) observe if a new Property called WorkshopCompleteDate was created, (3) why previousId was changed to previousIdId?, (4)
-

PR #502, opened 2021-07-30
Save workshop finishing date in the database

hnformentin:
Solves #474

### DIFF
--- a/backend/api/Services/EvaluationService.cs
+++ b/backend/api/Services/EvaluationService.cs
@@ -58,6 +58,24 @@ namespace api.Services
             return evaluation;
         }
 
+        public Evaluation SetWorkshopCompleteDate(Evaluation evaluation)
+        {
+            if (evaluation.WorkshopCompleteDate != null)
+            {
+                throw new InvalidOperationException(
+                    $"Completion date already set as: {evaluation.WorkshopCompleteDate}");
+            }
+            if (evaluation.Progression != Progression.FollowUp)
+            {
+                throw new InvalidOperationException(
+                    $"WorkshopCompleteDate requires an evaluation on FollowUp; it is: {evaluation.Progression}");
+            }
+            evaluation.WorkshopCompleteDate = DateTimeOffset.UtcNow;
+            _context.Evaluations.Update(evaluation);
+            _context.SaveChanges();
+            return evaluation;
+        }
+
         public IQueryable<Evaluation> GetAll()
         {
             return _context.Evaluations;
--- a/backend/tests/Services/EvaluationService.cs
+++ b/backend/tests/Services/EvaluationService.cs
@@ -88,5 +88,33 @@ namespace tests
             service.SetSummary(evaluation, summary);
             Assert.Equal(summary, evaluation.Summary);
         }
+        
+        [Fact]
+        public void SetWorkshopCompleteDate()
+        {
+            Project project           = GetProject();
+            EvaluationService service = new EvaluationService(_context);
+            Evaluation evaluation     = service.Create("eval_name", project, "");
+            System.DateTimeOffset testStartDate = System.DateTimeOffset.UtcNow;
+
+            Assert.Null(evaluation.WorkshopCompleteDate);
+
+            evaluation.Progression = Progression.Workshop;
+            var exceptionProgress = Assert.Throws<System.InvalidOperationException>
+                (() => service.SetWorkshopCompleteDate(evaluation));
+            Assert.Contains(
+                $"WorkshopCompleteDate requires an evaluation on FollowUp; it is: {evaluation.Progression}",
+                exceptionProgress.Message);
+
+            evaluation.Progression = Progression.FollowUp;
+            service.SetWorkshopCompleteDate(evaluation);
+            Assert.True(testStartDate < evaluation.WorkshopCompleteDate);
+
+            var exceptionPreviouslySet = Assert.Throws<System.InvalidOperationException>
+                (() => service.SetWorkshopCompleteDate(evaluation));
+            Assert.Contains(
+                $"Completion date already set as: {evaluation.WorkshopCompleteDate}",
+                exceptionPreviouslySet.Message);
+        }
     }
 }
--- a/frontend/src/api/models.ts
+++ b/frontend/src/api/models.ts
@@ -136,6 +136,21 @@ export type ComparableInt32OperationFilterInput = {
   nlte?: Maybe<Scalars['Int']>;
 };
 
+export type ComparableNullableOfDateTimeOffsetOperationFilterInput = {
+  eq?: Maybe<Scalars['DateTime']>;
+  neq?: Maybe<Scalars['DateTime']>;
+  in?: Maybe<Array<Maybe<Scalars['DateTime']>>>;
+  nin?: Maybe<Array<Maybe<Scalars['DateTime']>>>;
+  gt?: Maybe<Scalars['DateTime']>;
+  ngt?: Maybe<Scalars['DateTime']>;
+  gte?: Maybe<Scalars['DateTime']>;
+  ngte?: Maybe<Scalars['DateTime']>;
+  lt?: Maybe<Scalars['DateTime']>;
+  nlt?: Maybe<Scalars['DateTime']>;
+  lte?: Maybe<Scalars['DateTime']>;
+  nlte?: Maybe<Scalars['DateTime']>;
+};
+
 
 export type Evaluation = {
   __typename?: 'Evaluation';
@@ -149,6 +164,7 @@ export type Evaluation = {
   project: Project;
   summary?: Maybe<Scalars['String']>;
   previousEvaluationId?: Maybe<Scalars['String']>;
+  workshopCompleteDate?: Maybe<Scalars['DateTime']>;
 };
 
 export type EvaluationFilterInput = {
@@ -164,6 +180,7 @@ export type EvaluationFilterInput = {
   project?: Maybe<ProjectFilterInput>;
   summary?: Maybe<StringOperationFilterInput>;
   previousEvaluationId?: Maybe<StringOperationFilterInput>;
+  workshopCompleteDate?: Maybe<ComparableNullableOfDateTimeOffsetOperationFilterInput>;
 };
 
 export type GraphQuery = {
--- a/frontend/src/views/Evaluation/EvaluationGQL.ts
+++ b/frontend/src/views/Evaluation/EvaluationGQL.ts
@@ -21,7 +21,7 @@ export const useProgressEvaluationMutation = (): ProgressEvaluationMutationProps
     const apolloClient = useApolloClient()
 
     const PROGRESS_EVALUATION = gql`
-        mutation ProgessEvaluation($evaluationId: String!, $newProgression: Progression!) {
+        mutation ProgressEvaluation($evaluationId: String!, $newProgression: Progression!) {
             progressEvaluation(evaluationId: $evaluationId, newProgression: $newProgression) {
                 ...EvaluationFields
                 ...ParticipantsArray
--- a/frontend/src/views/Evaluation/FollowUp/Summaries/BowtieView.tsx
+++ b/frontend/src/views/Evaluation/FollowUp/Summaries/BowtieView.tsx
@@ -18,15 +18,21 @@ const ScrollableDiv = styled.div`
 
 interface Props {
     answersWithBarrier: AnswersWithBarrier[]
+    workshopCompleted?: string
 }
 
-const BowtieView = ({ answersWithBarrier }: Props) => {
+const BowtieView = ({ answersWithBarrier, workshopCompleted }: Props) => {
     return (
         <>
             <Box p="20px">
                 <Typography variant="h2" style={{ marginBottom: '30px', marginLeft: '30px' }}>
                     Bowtie model
                 </Typography>
+                {workshopCompleted && (
+                    <Typography style={{ marginTop: '-20px', marginLeft: '30px' }}>
+                        Workshop completed: {new Date(workshopCompleted).toLocaleDateString()}
+                    </Typography>
+                )}
                 <ScrollableDiv>
                     <BowtieWithBoxes answersWithBarrier={answersWithBarrier} />
                 </ScrollableDiv>
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -75,6 +75,7 @@ namespace api.GQL
 
             if (newProgression.Equals(Progression.FollowUp))
             {
+                _evaluationService.SetWorkshopCompleteDate(evaluation);
                 _answerService.CreateFollowUpAnswers(evaluation);
             }
 
--- a/backend/api/Migrations/20210811110505_AddWorkshopCompleteDate.Designer.cs
+++ b/backend/api/Migrations/20210811110505_AddWorkshopCompleteDate.Designer.cs
@@ -3,15 +3,17 @@ using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using api.Context;
 
 namespace api.Migrations
 {
     [DbContext(typeof(BmtDbContext))]
-    partial class BmtDbContextModelSnapshot : ModelSnapshot
+    [Migration("20210811110505_AddWorkshopCompleteDate")]
+    partial class AddWorkshopCompleteDate
     {
-        protected override void BuildModel(ModelBuilder modelBuilder)
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
--- a/backend/api/Migrations/20210811110505_AddWorkshopCompleteDate.cs
+++ b/backend/api/Migrations/20210811110505_AddWorkshopCompleteDate.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace api.Migrations
+{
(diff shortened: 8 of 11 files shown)


backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,21 @@ public void SetSummary()
...
+        public void SetWorkshopCompleteDate()
+        {
+            Project project           = GetProject();
+            EvaluationService service = new EvaluationService(_context);
+            Evaluation evaluation     = service.Create("eval_name", project);
+            DateTimeOffset testStartDate = DateTimeOffset.UtcNow;
+
+            Assert.Equal(null, evaluation.WorkshopCompleteDate);
+
+            // what is it missing here to set the end of an evaluation in a test?
```
hnformentin, line 102: How would it be the correct way to set that the evaluation is complete?

backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,21 @@ public void SetSummary()
             service.SetSummary(evaluation, summary);
             Assert.Equal(summary, evaluation.Summary);
         }
+
+        [Fact]
+        public void SetWorkshopCompleteDate()
```
hnformentin, line 93: For me some words like Fact and Assert  still appear underlined. Any tips on why this happens?

backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,21 @@ public void SetSummary()
...
+        {
+            Project project           = GetProject();
+            EvaluationService service = new EvaluationService(_context);
+            Evaluation evaluation     = service.Create("eval_name", project);
+            DateTimeOffset testStartDate = DateTimeOffset.UtcNow;
+
+            Assert.Equal(null, evaluation.WorkshopCompleteDate);
```
hnformentin, line 100: Is "null" the correct way to assert that evaluation.WorkshopCompleteDate was not yet set?

backend/tests/Services/EvaluationService.cs
hnformentin: it was related to the errors on the test I had wrote.

backend/tests/Services/EvaluationService.cs
hnformentin: Assert.Null is available. DateTimeOffset default value is MinValue instead of null, so the comparison was modified.

backend/tests/Services/EvaluationService.cs
hnformentin: no, the default value should be checked.

frontend/src/views/Evaluation/EvaluationGQL.ts
```diff
@@ -21,7 +21,7 @@ export const useProgressEvaluationMutation = (): ProgressEvaluationMutationProps
     const apolloClient = useApolloClient()
 
     const PROGRESS_EVALUATION = gql`
-        mutation ProgessEvaluation($evaluationId: String!, $newProgression: Progression!) {
+        mutation ProgressEvaluation($evaluationId: String!, $newProgression: Progression!) {
```
hnformentin, line 24: This is maybe a typo, How would it affect?

frontend/src/api/models.ts
```diff
@@ -1,608 +1,582 @@
...
+    id: Scalars['String']
+    name: Scalars['String']
+    createDate: Scalars['DateTime']
+    progression: Progression
+    status: Status
+    participants: Array<Maybe<Participant>>
+    questions: Array<Maybe<Question>>
+    project: Project
+    summary?: Maybe<Scalars['String']>
+    workshopCompleteDate?: Scalars['DateTime']
```
hnformentin, line 147: I only changed this line. What made everything to be updated?

hnformentin:
- How can I test that the modification in frontend models is working fine?
- How do I know if the variable should start with capital letter WorkshopCompleteDate (backend) or not workshopCompleteDate (frontend)?

hnformentin:
Is there a way to avoid showing the date if it is 01/01/1 (the default value)?

hnformentin:
This is when the evaluation is completed (correct):
Screenshot 2021-08-06 at 09 36 59

In this case we have the default value (the evaluation is not completed), but the default value is 01/01/1.
Screenshot 2021-08-06 at 09 36 40

frontend/src/views/Evaluation/EvaluationGQL.ts
denektenina: I think the mutation name doesn't matter, but I'm not sure

frontend/src/api/models.ts
denektenina: Do you use prettier for formatting maybe?

hnformentin: yes, so prettier would do all those changes?

denektenina: Prettier is usually configured to reformat on save (I don't know if you have to specify this yourself or if it just happens). So it might be that when you made a small change and then saved it reformatted the whole document.

hnformentin:
The nullable behaviour is implemented with the suggestion of @achaikou.

A test with non-complete evaluation:
Screenshot 2021-08-09 at 11 46 05 (2)


A test with completed evaluation:
Screenshot 2021-08-09 at 11 46 38 (2)

backend/api/Services/EvaluationService.cs
```diff
@@ -58,6 +59,20 @@ public Evaluation SetSummary(Evaluation evaluation, string summary)
             return evaluation;
         }
 
+        public Evaluation SetWorkshopCompleteDate(Evaluation evaluation)
+        {
+            bool isWorkshopComplete = (evaluation.Progression & Progression.FollowUp) == Progression.FollowUp;
+            evaluation.WorkshopCompleteDate = null;    
+
```
hnformentin, line 66: This way works, but I do not understand why do I need to set date WorkshopCompleteDate = null here and on the top when the evaluation was created. It would make sense for me if both were not necessary, e.g. the reference type would automatically be null.

frontend/src/api/models.ts
```diff
@@ -1,608 +1,582 @@
-export type Maybe<T> = T;
-export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
-export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
-export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
+export type Maybe<T> = T
```
achaikou, line 1: Oh, I think I have an idea what might cause such big differences!
It might be prettier or other default formatter.

when I run the `npm run schema` I have the version with semicolons.
If I additionally ask for prettier to format stuff (I do not have that done automatically, but may it be that you do?), all semicolons are gone and there are just a couple of changes left (I am not sure why they are present. I hope you didn't update `models.ts` manually? `npm run schema` does it for you on the basis of whatever you have in backend)

So if that's formatter, we just have to agree in the team whether we would have that file formatted by default with prettier, or leave it the way codegen generates it.

hnformentin:
Before merging, I plan to rebase and remove the modifications in the commit "Copy modifs from equinor#430". This was only used to be able to test.

achaikou (commented):
Seems good from my side! :cat: 
Will be approving after all the comments are discussed/rejected/agreed and all the commit-rebase stuff done.

Additional comments:

You might also would want to have just two commits:
1. Fixing typo
2. Doing all the complete date stuff. (In commit 4 you remove some things you introduce in commit 1)

I also assume that it was discussed how all is supposed to work and where this workshop date is supposed to be placed, as original issue lacks information.

backend/api/Services/EvaluationService.cs
```diff
@@ -58,6 +58,17 @@ public Evaluation SetSummary(Evaluation evaluation, string summary)
             return evaluation;
         }
 
+        public Evaluation SetWorkshopCompleteDate(Evaluation evaluation)
+        {
+            if (evaluation.Progression == Progression.FollowUp)
```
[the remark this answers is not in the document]
hnformentin: This is good reasoning. I will try.

backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,20 @@ public void SetSummary()
...
+            EvaluationService service = new EvaluationService(_context);
+            Evaluation evaluation     = service.Create("eval_name", project, "");
+            System.DateTimeOffset testStartDate = System.DateTimeOffset.UtcNow;
+
+            Assert.Equal(null, evaluation.WorkshopCompleteDate);
+
+            evaluation.Progression = Progression.FollowUp;
+            service.SetWorkshopCompleteDate(evaluation);
+            Assert.True(testStartDate < evaluation.WorkshopCompleteDate);
+        }
```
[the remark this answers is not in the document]
hnformentin: Good suggestion, it makes sense to me. I will try to do it.

frontend/src/views/Evaluation/FollowUp/Summaries/BowtieView.tsx
```diff
@@ -18,15 +18,21 @@ const ScrollableDiv = styled.div`
...
         <>
             <Box p="20px">
                 <Typography variant="h2" style={{ marginBottom: '30px', marginLeft: '30px' }}>
                     Bowtie model
                 </Typography>
+                {workshopCompleted && (
+                    <Typography style={{ marginTop: '-20px', marginLeft: '30px' }}>
+                        Workshop completed: {new Date(workshopCompleted).toLocaleDateString()}
+                    </Typography>
+                )}
```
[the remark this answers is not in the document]
hnformentin: I would love to start using cypress. This frontend code was taken from a previous issue #430 . Depending on what we conclude for #430 (should be together with this or on the separate issue), this can be done here or there.

hnformentin:
It also misses the migration files. Check Erlend's PR and the Readme.md.

backend/tests/Services/EvaluationService.cs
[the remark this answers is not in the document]
hnformentin: @achaikou I am trying the assertion below but having two problems in the tests so far:

`Assert.Throws<InvalidOperationException>(() => service.SetWorkshopCompleteDate(evaluation)($"To set WorkshopCompleteDate, the evaluation should be on FollowUp progress, but it is in : {evaluation.Progression}"));`

(1) The type or namespace name 'InvalidOperationException' could not be found (are you missing a using directive or an assembly reference?) [tests]
(2) Evaluation EvaluationService.SetWorkshopCompleteDate(Evaluation evaluation) Method name expected [tests]
Do you have any idea about that?

I am not being able to get on those, if you have suggestions highly welcome.

frontend/src/views/Evaluation/EvaluationGQL.ts
hnformentin: I will correct it as a typo. It seems not to make a difference indeed.

backend/api/Services/EvaluationService.cs
```diff
@@ -58,6 +58,29 @@ public Evaluation SetSummary(Evaluation evaluation, string summary)
...
+            {
+                evaluation.WorkshopCompleteDate = DateTimeOffset.UtcNow;
+                _context.Evaluations.Update(evaluation);
+                _context.SaveChanges();
+                return evaluation;
+            }
+            else
+            {
+                throw new NotImplementedException($"None of the conditions to set the WorkshopCompleteDate are met.");    
+            }
```
hnformentin, line 81: I added this else statement because there was an warning about having paths that did not return anything.

backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,27 @@ public void SetSummary()
...
+        {
+            Project project           = GetProject();
+            EvaluationService service = new EvaluationService(_context);
+            Evaluation evaluation     = service.Create("eval_name", project, "");
+            System.DateTimeOffset testStartDate = System.DateTimeOffset.UtcNow;
+
+            Assert.Null(evaluation.WorkshopCompleteDate);
+
+            evaluation.Progression = Progression.Workshop;
+            var ExceptionProgress = Assert.Throws<System.InvalidOperationException>(() => service.SetWorkshopCompleteDate(evaluation));
```
achaikou, line 103: `ExceptionProgress` and `ExceptionPreviouslySet` are local variable, so as per C# naming conventions their names [must start from the lower case](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables) - `exceptionProgress `, `exceptionPreviouslySet`
(I actually can't recall any language where local variable would start from Upper case, so that's a very, very common convention)

backend/tests/Services/EvaluationService.cs
```diff
@@ -88,5 +88,27 @@ public void SetSummary()
             service.SetSummary(evaluation, summary);
             Assert.Equal(summary, evaluation.Summary);
         }
```
achaikou, line 90: P.S.
You might want to wish to format this file. Now some lines (yours and Erlends from the last change) are changed on formatting

~Speaking of Erlend: Erlend would have complained about too long lines here. I kind of agree, but I am not Erlend, so I will leave this pleasure to him whenever he happens to do the code review for you~ :smile:

backend/api/Services/EvaluationService.cs
hnformentin: It is simpler indeed! Why not use `else if` instead of the second if? I checked a couple of discussions; it seemed to me that in this case where we are throwing exceptions for each if, it would make sense to use else-if. But maybe you have a different reasoning?

achaikou: Mostly style? Which means that you kinda can do what you find most readable?

I would say everything works fine without second else, so we can ignore it.
If you have two conditions that are somewhat connected it would make more sense to write something like that.

```
if (directory.notFound) {
    throw DirectoryNotFoundOrSomething()
} else if (directory.isEmpty) {
   throw SomeOtherError("we did not expect directory to be empty")
} else {
  doMyUsefulJob(directory)
}
```

Our conditions have nothing to do with each other, so why introduce additional "else"?

backend/tests/Services/EvaluationService.cs
achaikou: :smiling_imp: 
Here there are no global conventions, as many people disagree with each other, so it is usually decided team by team.
Last time Erlend tried to introduce limit of 80 characters per line. He lost :) 

In the frontend limit is set to 140 characters, but there is no explicitly defined limit in C#.
I would say that if you see that code doesn't fit on screen or if your introduced lines are longer than already existing in the file, it might be a good idea to add some breaks.

backend/api/Services/EvaluationService.cs
hnformentin: Good point Alena. For my understanding, it would make sense to use else-if in two situations: (1) when two conditions are connected (as your example), (2) when the statements return exceptions (to indicate the second condition doesn't need to be checked if the first is true - the current situation). So the code would be very similar to your suggestion:
`            if (evaluation.WorkshopCompleteDate != null)
            {
                throw new InvalidOperationException($"The completion date of the evaluation is already set as: {evaluation.WorkshopCompleteDate}");
            }
            else if (evaluation.Progression != Progression.FollowUp)
            {
                throw new InvalidOperationException($"To set WorkshopCompleteDate, the evaluation should be on FollowUp progress, but it is in: {evaluation.Progression}");
            }

            evaluation.WorkshopCompleteDate = DateTimeOffset.UtcNow;
            _context.Evaluations.Update(evaluation);
            _context.SaveChanges();
            return evaluation;`
            
But I am fine to adopt what would be more natural for this project, no strong opinions :-D

hnformentin:
A new commit addressing Alena's review is there. 
If the modifications are fine, I plan to rebase only with the three previous commits.
I guess it is not necessary to do the migration again, since the changes where not in the database fields. Is it correct?

achaikou:
OK, feel free to do clean rebase, will approve right after :+1: . I trust Kristian that migrations are good :)


1. One small request for future: please don't resolve issues started by me. (Don't know what other think about that).
It makes more sense for reviewer to see if their changes were addressed as intended/acceptable and resolve them themself.

2. Your fixups as "Alena's review" are fine, but if you have several commits then you can use pure [fixups](https://stackoverflow.com/a/3828861).
(should also be easily available in any GIT-GUI: choose commit - click fixup - done.
When you rebase, they will automatically apply to the commit they refer to, without manual step of squashing.)

3. Also not sure why you want 3 commits, not 2. Do you believe migrations are not part of your main commit?

4. Small addition to our "exceptions" discussion (I do not expect any changes with regards to it)
Apparently C# has something called Contracts.
It allows writing checks aka [preconditions](https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/code-contracts#preconditions) in even better way 
```

  Contract.Requires(evaluation.WorkshopCompleteDate == null);
  Contract.Requires(evaluation.Progression == Progression.FollowUp, "here goes the message if we want to");

  evaluation.WorkshopCompleteDate = DateTimeOffset.UtcNow;
```

However they seem to require [additional modules](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.contracts.contract?view=net-5.0#remarks) to actually check stuff in runtime and doesn't seem to be worth dealing with it as `if-throw` works well.

5. Never let me sit too long on the review, I will find new things to complain about... :facepalm:

Small thing (and not for this PR) : we now have 2 different formats in the application

- 1) your changes print date in "local" datetime format (which I believe is good). I don't know what format I have on my machine, but it happens to show the difference:
`Workshop completed: 8/12/2021`
On Follow up/Actions tab Due date is displayed in the same format: `8/12/2021 `
- 2) But if you do Add Action (or Edit action) - and see date in the calendar chooser, it would be `12/08/2021`.
This is caused by [fusion enforcing this format](https://github.com/equinor/fusion-api/blob/b5629c0195606c23ee5d8ea1feba9d6a28f93830/src/intl/DateTime.ts#L70)

Now it is kinda fine, as we use fusion date only in the date picker, but if we have in the future some bigger mix of fusion/local dates this can lead to certain confusion.
We might want to mention it to Nina and figure out if we want consistency.

achaikou (approved):
:tada:

hnformentin:
Closes #430

PR #507, opened 2021-08-04
Progress indicator in workshop tab now counts all answers

denektenina:
**Context**: In all steps of BMT, there is a side panel next to the questions that count how many questions are answered. The Workshop tab works a bit different than the other tabs, because here only facilitators can write answers, and if there are more than one facilitator, there can be answers from either of these on either of the questions.

image

**Problem**: Even though multiple respondents can answer the questions, in the side panel, only the answers written by the logged in user will be counted. Therefore, we can easily get a situation where you can see that 10 out of 37 questions is answered, but in the side panel it only says that 1/37 questions are answered.

**Solution**: For the Workshop tab (`Progression.Workshop` in the code), we have to count all answers, not only the answers that match the logged in users azureId.

This PR looks like there is more new code than t here is because of prettier formatting, the only real change is the following: 

`const participantAnswers =
        progression === Progression.Workshop
            ? progressionAnswers
            : progressionAnswers.filter(a => a.answeredBy?.azureUniqueId === azureUniqueId)`

I.e. we now check if we are on the workshop step and only filter on azure ID if we aren't. Earlier code:

`const participantAnswers = progressionAnswers.filter(a => a.answeredBy?.azureUniqueId === azureUniqueId)`

### DIFF
--- a/frontend/src/utils/QuestionAndAnswerUtils.test.ts
+++ b/frontend/src/utils/QuestionAndAnswerUtils.test.ts
@@ -28,84 +28,178 @@ describe('Test QuestionAndAnswerUtils', () => {
 
             expect(isFilledOut).toBe(true)
         })
-    it('Check if getFilledUserAnswersForProgression gets correct answers', () => {
-        const evaluation: Evaluation = {
-            createDate: new Date(),
-            id: '',
-            name: '',
-            participants: [],
-            progression: Progression.Preparation,
-            status: Status.Active,
-            project: {
-                createDate: new Date(),
-                evaluations: [],
-                fusionProjectId: '',
-                id: '',
-            },
-            questions: [],
-        }
+})
 
-        const question: Question = {
-            actions: [],
-            answers: [],
-            barrier: Barrier.Gm,
-            createDate: new Date(),
-            evaluation,
-            order: 1,
-            id: '',
-            organization: Organization.All,
-            questionTemplate: {
-                barrier: Barrier.Gm,
-                createDate: new Date(),
-                id: '',
-                order: 1,
-                organization: Organization.All,
-                questions: [],
-                status: Status.Active,
-                supportNotes: '',
-                text: '',
-            },
-            supportNotes: '',
-            text: '',
-        }
-
-        const participant: Participant = {
-            azureUniqueId: '1',
-            createDate: new Date(),
-            evaluation,
-            evaluationId: evaluation.id,
-            id: '',
-            organization: Organization.All,
-            progression: Progression.Preparation,
-            role: Role.OrganizationLead,
-        }
-
-        const answer1: Answer = {
-            createDate: new Date(),
-            id: '',
-            progression: Progression.Preparation,
-            question,
-            questionId: '',
-            severity: Severity.High,
-            text: 'answer1',
-            answeredBy: participant,
-        }
-
-        const answer2: Answer = {
-            createDate: new Date(),
-            id: '',
-            progression: Progression.Individual,
-            question,
-            questionId: '',
-            severity: Severity.High,
-            text: 'answer2',
-            answeredBy: participant,
-        }
-
+describe('Test getFilledUserAnswersForProgression', () => {
+    it('gets answers only from the stage the user is currently at', () => {
         question.answers = [answer1, answer2]
 
         const answers = getFilledUserAnswersForProgression([question], Progression.Preparation, participant.azureUniqueId)
 
         expect(answers.length).toBe(1)
     })
+
+    it('get only the users own answers, not answers from others, from the current stage', () => {
+        question.answers = [answer2, answer5]
+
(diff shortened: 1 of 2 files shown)


frontend/src/utils/QuestionAndAnswerUtils.test.ts
```diff
@@ -28,84 +28,170 @@ describe('Test QuestionAndAnswerUtils', () => {
...
         })
-    it('Check if getFilledUserAnswersForProgression gets correct answers', () => {
-        const evaluation: Evaluation = {
+})
+
+describe('Test getFilledUserAnswersForProgression', () => {
+    it('gets answers for preparation progression on preparation step', () => {
+        question.answers = [answer1, answer2]
+
+        const answers = getFilledUserAnswersForProgression([question], Progression.Preparation, participant.azureUniqueId)
```
achaikou, line 37: I see that this test was not written in this commit, so just asking then: do you maybe know what behavior it was supposed to test?
answer 1 and 2 are both given by the same user. Test is supposed to tell me something about preparation step.
I am confused:
- if the test was just supposed to test that answers for Individual are not visible on Preparation, then test description is confusing
- if test was supposed to test preparation step on its own, then why just 1 user? Question should have answer in Progression stage by different user as well...

frontend/src/utils/QuestionAndAnswerUtils.test.ts
```diff
@@ -28,84 +28,170 @@ describe('Test QuestionAndAnswerUtils', () => {
...
+
+        expect(answers.length).toBe(1)
+        expect(answers[0].answeredBy && answers[0].answeredBy.azureUniqueId).toBe(participant.azureUniqueId)
+    })
+
+    it('gets answers from all users on workshop step', () => {
+        question.answers = [answer3, answer4]
+
+        const answers = getFilledUserAnswersForProgression([question], Progression.Workshop, facilitator1.azureUniqueId)
+
```
achaikou, line 55: One other cool test would be to check that not only facilitator, but a random participant can see correct number of answers. I believe they would get 0 before and correct number now.

achaikou (approved):
Thanks for explanations! Knew what to look at! :heart: 

Code looks good. See if you want to update the tests, if not - just merge.
Tested that it works as intended using Cypress setups and mock users, but I hope we will find a better way than the one I tried...

frontend/src/utils/QuestionAndAnswerUtils.test.ts
denektenina: The test description is the following: "it gets answers for **preparation** progression on **preparation** step" - i.e. given two questions by the same user but on different steps (you can see that answer one has: `progression: Progression.Preparation`, and answer2 has `progression: Progression.Individual`) you only get the answer from the correct step (here: prepraration). The test checks that gicen two questions, one is filtered out and one is returned.

To answer your ponints:
* Do you understand the test description better now? I can maybe change it to "it ONLY gets answers from preparation progression...."?
* I think we test that it filters out other users in later test?

achaikou: OK, I see the problem.

Judging the all the present test, the description I was expecting to see is
"user gets own answers only from the stage they are currently at"

If we formulate it like that it doesn't really matter if the stage is preparation or for example workshop as user on any stage is supposed to get answers from the different stage from the one they are at.

If stages matter, we might want to have tests "user in progression gets own answers only from progression stage"


And the next test is actually "user gets only their own answers, not answers from others, from the current stage". 
Test is done only on Individual step. But it can also be a Preparation step, right?

So what I actually mean in this case that if we have "individual"/"preparation" stage in the test title, it means that stage is somehow important for the test and raises the question why other stages are not tested for the same behavior, like " user does not get answers from individual stage on workshop" or "user gets only their own answers in preparation stage" or "who knows what happens on follow-up stage"?

If stage is not explicitly mentioned, it means that we do not care at what stage user is, test is valid at any stage, and Individual/Preparation are just examples.

frontend/src/utils/QuestionAndAnswerUtils.test.ts
achaikou: I think we can cover it with a simple unit test, no?
Just have the same setup as now, but instead of `facilitator1.azureUniqueId` put `participant.azureUniqueId`?

I checked that yesterday, participant does see the full answers number of workshop questions on the sidebar.

frontend/src/utils/QuestionAndAnswerUtils.test.ts
denektenina: true, I understand what you mean. I can change that for clarity.

frontend/src/utils/QuestionAndAnswerUtils.test.ts
denektenina: Ah ok, yes, I can make a test like that

PR #508, opened 2021-08-04
Stabilize page visit

achaikou:
Apparently last attempt to solve it didn't manage to do it completely.
Here goes new attempt:

1. Before waiting for "Loading" to disappear check if we are not on
Evaluations page already. It looks like in rare cases page is loaded
before we start checking for "Loading" sign.

2. Simulate fusion (again) by saving project information in cache. That
seems to solve the problem of project not being loaded on navigation and
"Please select a project" label persisting.

### DIFF
--- a/frontend/cypress/support/visit.ts
+++ b/frontend/cypress/support/visit.ts
@@ -5,13 +5,18 @@ function setupEnvironment() {
     cy.viewport(1800, 1000) //until we decide to test on various resolutions
 }
 
-function waitForFusionPageLoad() {
-    cy.contains('Please select a project.').should('exist')
-}
-
 function waitForProjectPageLoad() {
-    cy.contains('Loading').should('exist')
-    cy.contains('Loading', { timeout: 10000 }).should('not.exist')
+    cy.get('body').then(body => {
+        /**
+         * Two cases possible: either we already are on the Project page
+         * and "Loading" element is long gone, or page didn't even start
+         * loading and element is not there yet.
+         */
+        if (body.find('div:contains(Evaluation)').length === 0) {
+            cy.contains('Loading', { timeout: 10000 }).should('exist')
+        }
+        cy.contains('Loading', { timeout: 20000 }).should('not.exist')
+    })
 }
 
 function waitForEvaluationPageLoad() {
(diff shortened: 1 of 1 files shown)


frontend/cypress/support/visit.ts
```diff
@@ -5,13 +5,18 @@ function setupEnvironment() {
...
-    cy.contains('Loading').should('exist')
-    cy.contains('Loading', { timeout: 10000 }).should('not.exist')
+    cy.get('body').then(body => {
+        /**
+         * Two cases possible: either we already are on the Project page
+         * and "Loading" element is long gone, or page didn't even start
+         * loading and element is not there yet.
+         */
+        if (body.find('div:contains(Evaluation)').length == 0) {
+            cy.contains('Loading').should('exist')
```
ReedOnly, line 16: should have a timeout for this one as well?

frontend/cypress/support/visit.ts
```diff
@@ -5,13 +5,18 @@ function setupEnvironment() {
...
 function waitForProjectPageLoad() {
-    cy.contains('Loading').should('exist')
-    cy.contains('Loading', { timeout: 10000 }).should('not.exist')
+    cy.get('body').then(body => {
+        /**
+         * Two cases possible: either we already are on the Project page
+         * and "Loading" element is long gone, or page didn't even start
+         * loading and element is not there yet.
+         */
+        if (body.find('div:contains(Evaluation)').length == 0) {
```
ReedOnly, line 15: what Codacy says

ReedOnly (changes requested):
It was only 21 hours ago I was assigned ⏲️

frontend/cypress/support/visit.ts
achaikou: At the moment it has a default timeout of 4 seconds and that sufficed in all local runs.
We don't know what's going to happen in CI, so we don't know how possible it is that between end of visit and Loading label appearing >4 seconds will pass.
In general they do not recommend setting too long waits because it causes test to fail way later than it should.

But ok, given the CI slowness lets increase the timeout from default 4 to custom 10 seconds and see how well it will work out? Setup you created is perfect, so if there is any failure related to this ever, we would be able to adjust accordingly.

achaikou:
Good! I was not surprised you didn't review it, I was just surprised you didn't know you've been assigned given that we spoke about this PR recently :stuck_out_tongue:  :smiley:

ReedOnly (approved):

PR #509, opened 2021-08-05
Cypress users

achaikou:
Two previously wanted changes:
- instead of always writing `users[0] `or similar, allow randomization of users. We do not currently distinguish between user roles, but as we are dependent on fusion (and apparently there there are some issues with users roles) it might be good to check that we do not provide any new and unexpected problems. (Note that we won't be able to detect real fusion issues with these users, as we mock fusion responses as much as we can)
- save tokens to avoid querying oath2 sever every time

Merge of this PR can wait until Cypress CI PR is merged.


ReedOnly (approved):
Remember to rebase!

achaikou:
Question:
I want to test these changes on the cypress branch. But.
I am kinda scared of accidentally pushing (and force-pushing, as we use it often) stuff into the wrong branch in upstream. So I have pushing disabled for upstream.
If I want to enable it, then can we maybe have a branch protection in github? We have one, but it "include administrators" is not checked, so I am not sure if us who have full access to this repo won't be able to do it anyway.

I start thinking that better option might be to directly enable cypress tests on PRs. They will take long, true, but it removes a need for an important manual step.

PR #510, opened 2021-08-05
Facilitator can now remove facilitator

denektenina:


frontend/cypress/integration/nominations_spec.ts
```diff
@@ -0,0 +1,44 @@
...
+                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
+                cy.contains(owner.user.name).should('exist')
+                cy.contains(otherFacilitator.user.name).should('exist')
+            })
+        })
+
+        it('Facilitator can delete an other facilitator', () => {
+            const nominationPage = new NominationPage()
+            nominationPage.deletePerson(otherFacilitator.user)
+            cy.testCacheAndDB(() => {
```
achaikou, line 25: Didn't review yet, but know what the problem is. That was my mistake when I created this function, but why it worked without it from actions_spec - no idea.

Basically you would need to put
`import './common'`
into `support/index.ts` and everything works

denektenina: Great! Nice catch!

frontend/cypress/support/nomination.ts
```diff
@@ -23,6 +33,12 @@ export default class NominationPage {
         })
     }
 
+    assertParticipantNotThere = (user: User) => {
```
achaikou, line 36: I recall that we came up with this name quite fast, but I am not sure it's the best one.
Maybe we should rename these two methods to `assertParticipantPresent` and `assertParticipantAbsent `?

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -0,0 +1,61 @@
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { Progression, Role } from '../../src/api/models'
+import { Participant } from '../support/mocks'
+import NominationPage from '../support/nomination'
+import users from '../support/users'
```
achaikou, line 5: Can be deleted

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            cy.contains(otherFacilitator.user.name).should('exist')
+
+            const nominationPage = new NominationPage()
+            nominationPage.deletePersonDiv(owner.user).should('not.exist')
+            nominationPage.deletePerson(otherFacilitator.user)
+            cy.testCacheAndDB(() => {
+                nominationPage.assertParticipantNotThere(otherFacilitator.user)
+            })
+        })
+
```
achaikou, line 31: I am not very fond of duplication, and both out tests duplicate.
What about creating a separate function?

 ```
       function checkParticipantCanBeDeleted(toDelete: Participant, toStay: Participant) { //or users instead of participants as we anyway always use the User field 
            cy.visitEvaluation(seed.evaluationId, toStay.user)
             ...
            cy.visitProject(toDelete.user)
            cy.contains(seed.name).should('not.exist')
        }

        it('Facilitator can delete an other facilitator, but not themself', () => {
            checkParticipantCanBeDeleted(otherFacilitator, owner) //or any other cooler name for this function
        })

        it('Other facilitator who is not owner can delete owner', () => {
            checkParticipantCanBeDeleted(owner, otherFacilitator)
        })
```

**Review by achaikou — APPROVED** [review]
Looks good to me!
See if you want to fix any my or Codacy comments.

Approving as I see nothing that I can't live without :)

File: frontend/src/views/Evaluation/Nomination/NominationTable.tsx [hunk 5]
```diff
@@ -23,19 +24,23 @@ interface DataTableRowProps {
...
-            Delete
-        </Button>
+        <div data-testid={'delete_button_' + item.participant.azureUniqueId}>
+            <Button
+                onClick={() => {
+                    deleteParticipant(item.participant.id)
+                }}
+                disabled={item.progressionStatus !== ProgressionStatus.InProgress || loading}
+            >
+                Delete
```
achaikou, line 41: Unlikely to happen, but we might need to try to defend against it:
facilitators A and B in two browsers.
A removes B.
B removes A.

There are no more facilitators left.

denektenina: Yes, but this probably goes under the new task, where we defend against users deleting where they are not allowed? I mean, they can only remove each other like that if we allow non-facilitators to remove facilitators, and we haven't decided if that should be allowed?

frontend/cypress/integration/nominations_spec.ts
denektenina: Created two separate functions with shared code.

frontend/cypress/integration/nominations_spec.ts
denektenina: Fixed

frontend/cypress/support/nomination.ts
denektenina: Yes, changed the names

frontend/cypress/integration/nominations_spec.ts
denektenina: And re-ran the tests to ensure it still works

frontend/src/views/Evaluation/Nomination/NominationTable.tsx
achaikou: I see what you mean. We should have this as a validation in the backend that checks requester abilities.
You are right. Different issue.

achaikou (approved):
Looks good to me!
Squash and merge :)

PR #513, opened 2021-08-09
Disable backend authorization for dev environments

### LINKED ISSUE #484: BMT: Simplify how to switch off backend authorization

achaikou:
Having to manually update "Require authorization" fields is
1) inconvenient
2) not exactly safe if someone unintentionally commits it and reviewer
doesn't notice

closes #484 

Upd: the original version of this PR didn't fail any of the used checks despite having a big problem for Production/Test code.
It should have failed Cypres tests, however.
It might be that running all Cypress tests for every PR might be desirable.

### DIFF
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -199,21 +199,35 @@ namespace api
                     { { "resource", $"{Configuration["AzureAd:ClientId"]}" } });
             });
 
-            app.UseAuthorization();
-
-            /* Run locally without Authorization
-             *
-             * When developing locally it can useful to disable authorization
-             * of the exposed endpoints. This can be done by temporarily
-             * removing 'RequireAuthorization()` for MapGraphQL and
-             * MapControllers.
-             */
-            app.UseEndpoints(endpoints =>
+            if (env.IsDevelopment())
             {
-                endpoints.MapHealthChecks("/health").AllowAnonymous();
-                endpoints.MapGraphQL().RequireAuthorization();
-                endpoints.MapControllers().RequireAuthorization();
-            });
+                /* Disable endpoints authorization for development environment
+                 *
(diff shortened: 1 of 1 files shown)


ReedOnly (changes requested):
In theory I like your suggested changes, however it has a few major concerns:
1. We don't want to have a deploy of fusion anywhere in radix without authentication, which it will now for master branch
2. If someone somehow intentionally or accidentally changes the ASPNETCORE_ENVIRONMENT env variable to Development in Radoxconfig suddenly all data will be exposed
3. We want authentication to be enables as a default, also for local development. For many local uses cases you don't have to disable authentication

achaikou:
Remember that in the worst case where we screw up this is exactly the behavior we had before the 6de93295dce491eb5a4ffa61941022987b0efe6d meaning that only endpoints would be exposed.

1. Then the question is why do we deploy anywhere to radix in development mode? Shouldn't we use test/production mode?
2. I think that's a default danger of someone doing something stupid when setting up production environment? Simply do not provide env variable in Radixconfig and all will be fine as value would be [production](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-5.0#set-the-environment)?
3. Would depend on what you want. Only one situation I can think off are unauthorized messages when listing evaluations.

You know what? Would it be satisfactory to you if we provided a separate environment variable ENDPOINTS_AUTHORIZATION instead of relying on `env.isDevelopment()`? If it's not explicitly set to False then True value will hold. If it is explicitly set, then endpoints authorization will be turned off.
Is that acceptable for you?

achaikou:
It has been decided that having environment variable to disable authorization is more dangerous than commenting the code out as it is possible to do in production and not notice.
Hence closing

PR #529, opened 2021-08-16
Show project name in single user action table

maiaht:


ReedOnly:
looks good, but does the project name display later than the other fields?

frontend/src/components/ActionTable/ActionTable.tsx
```diff
@@ -139,6 +148,7 @@ const ActionTable = ({ onClickAction, actionsWithAdditionalInfo, personDetailsLi
                         const priority = action.priority
                         const priorityFormatted = priority.substring(0, 1) + priority.substring(1).toLowerCase()
                         const assignedTo = assignedPersonDetails(action)
+                        const projectName = getFusionProjectName(projects, action.question.evaluation.project.fusionProjectId)
```
ErlendHaa, line 151: This apparently causes an issue in some cases. And it's an interesting one. I.e. we have cases where `question` is undefined on the `action`.

The situation in which `question` is undefined happens when we navigate to an existing evaluation, which is at stage Follow-up. (But only if there exists any actions). The effect is that the entire evaluation is un-viewable

To explain in detail what's going on:

Consider you navigate to an evaluation, e.g. `localhost:3000/projectID/evaluation/evaluationId`. At that point, [this query](https://github.com/equinor/fusion-bmt/blob/0ffdd652f5ec86d9eb00cff1b85fdae134951eac/frontend/src/views/Evaluation/EvaluationGQL.ts#L96) is executed and then the page loads. Notice how the Answer fragment in that query does _not_ include `Answer.question`. Then if the Evaluation is in follow-up, `ActionTable` will be loaded - and crashes on the undefined `Action.question` on this line

maiaht:
> looks good, but does the project name display later than the other fields?

This has been fixed now

frontend/src/components/ActionTable/ActionTable.tsx
```diff
@@ -50,10 +52,12 @@ interface Props {
     actionsWithAdditionalInfo: ActionWithAdditionalInfo[]
     personDetailsList: PersonDetails[]
     onClickAction: (action: Action) => void
-    singleUser?: boolean
+    evaluations?: Evaluation[]
```
ErlendHaa, line 55: I still think this should be a boolean, because it servers simply as switch to whether the Evaluation column is included or not. No data is used from the currently passed Evaluation Array

maiaht: Fixed

ReedOnly (approved):

PR #531, opened 2021-08-16
Add role-based access to answers

### LINKED ISSUE #470: Only Facilitators and org lead should be able to view individual answers in answer sidebar
We do this change to allow for opening up the tool to everyone in Equinor

ErlendHaa:
This commit prohibits participants with role Participant or ReadOnly to
view other participants individual answers in the AnswerSummarySidebar.

Closes #470


ReedOnly (approved):
Looks good, but we'll the role base access in #537

PR #532, opened 2021-08-16
Add cypress tests for actions

### LINKED ISSUE #505: Add create/edit Actions Cypress tests

achaikou:
closes #505


frontend/cypress/support/evaluation.ts
```diff
@@ -0,0 +1,56 @@
+import { Progression } from '../../src/api/models'
+
+export class EvaluationPage {
+    progressionStepLink = (progression: Progression) => {
+        let progressionText: string
+        switch (progression) {
```
achaikou, line 6: As this is used for navigation only, `EnumToString.progressionToString` can be used instead

ErlendHaa: `EnumToString.progressionToString` will yield "FollowUp", not "Follow-up". But should work for all other cases. In the interest of keeping the method short, consider reworking the `switch` to use fall-through for all but Followup, either with `default` or explicit fall-through for each Progression:

```
switch(progression) { 
    case Progression.FollowUp: {
        progressionText = 'Follow-up'
        break
    }
    default: {
        progressionText = EnumToString ...
    }
}

File: frontend/cypress/integration/actions_spec.ts [hunk 2]
```diff
@@ -75,11 +269,110 @@ describe('Actions', () => {
...
+                    .should('have.text', a.completed ? 'Yes' : 'No')
+                cy.get('@row').eq(4).should('have.text', mapPriority(a.priority))
+                cy.get('@row').eq(5).should('have.text', a.assignedTo.user.name)
+                cy.get('@row').eq(6).should('have.text', a.dueDate.toLocaleDateString())
+            })
+        })
+    })
 })
 
+const createCreateSeed = () => {
```
achaikou, line 333: TODO for all seeds:
I again forgot that in current solution user's role is not randomized by default and is always default static.
Change to all possible.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -1,48 +1,240 @@
...
+        switch (priority) {
+            case Priority.Low:
+                return 'Low'
+            case Priority.Medium:
+                return 'Medium'
+            case Priority.High:
+                return 'High'
+        }
+    }
+
```
achaikou, line 36: Structure its better.
At the moment we have two equivalence classes for modification (create/edit/delete) : on Workshop/Follow up and on Nomination/Individual/Preparation. We test only one of them. Test that no actions are possible on the Nomination/Individual/Preparation.

Once we have a distinction between viewers/people who can do stuff, we would also test that viewers can't do delete/add/edit

frontend/cypress/support/external.ts
```diff
@@ -82,3 +82,27 @@ declare global {
...
+    if (err.message.includes("Cannot read property 'removeEventListener' of null")) {
+        return false
+    }
+
+    /* might get thrown if intercepted fusion calls were interrupted,
+     * for example by new test start */
+    const unauthorized = /\[https:\/\/pro\-s\-people\-ci\.azurewebsites\.net.+] returned HTTP status code \[401\]/
+    if (err.message.match(unauthorized)) {
+        return false
+    }
```
achaikou, line 107: This one might not be needed after all.
It most likely falls under unhandled promises?
Though this was a one-time error message from Petter's machine, so not 100% sure.

frontend/cypress/integration/actions_spec.ts
petterwildhagen: These tests can be added to the test-levels of what is possible/not possible instead of the feature-level.

frontend/cypress/support/evaluation.ts
ErlendHaa: Disregard my above comment, I did not realise we implement our own EnumToString thing

frontend/cypress/integration/actions_spec.ts
```diff
@@ -1,48 +1,240 @@
...
+                cy.visitEvaluation(seed.evaluationId, editor.user)
+                evaluationPage.progressionStepLink(actionProgression).click()
+            })
+        })
+
+        /**
+         * Check provided values on the Edit dialog.
+         * @param notes Pass all notes in the order of creation
+         */
+        function checkActionValues(
```
petterwildhagen, line 113: Move into class ActionDialog

petterwildhagen: Pass entire object of type Action as argument?, i.e. 
funciton checkActionValue(action: Action, notes: { participant: Participant; note: string }[])

frontend/cypress/integration/actions_spec.ts
```diff
@@ -1,48 +1,240 @@
...
+            const actionTitle = 'Feel proud for me, I have been updated!'
+            const assignee = faker.random.arrayElement(seed.participants)
+            const dueDate = faker.date.future()
+            const priority = mapPriority(faker.random.arrayElement(Object.values(Priority)))
+            const description = faker.lorem.words()
+            const completed = !existingAction.completed
+            const onHold = !existingAction.onHold
+            const notes: string[] = Array.from({ length: faker.datatype.number({ min: 1, max: 3 }) }, () => {
+                return faker.lorem.words()
+            })
```
achaikou, line 148: As per suggestion from Petter, move this and similar in other test to something like createEditTestData (see if it is possible to reuse Action class for return values - unclear).
This will allow us to make test a bit smaller

achaikou: That was an awesome idea, Petter! I like the design way more now :heart:

frontend/cypress/integration/actions_spec.ts
achaikou: Updated:
When discussing it last week I believe we agreed to have these as view/business-logic tests, so not implementing anything here.
What can/cannot be done by whom and where would be implemented in different tests suite.
This one will test features themselves.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -1,48 +1,240 @@
...
+            ;({ seed } = createCreateSeed())
+
+            seed.plant().then(() => {
+                const user = faker.random.arrayElement(seed.participants).user
+                cy.visitEvaluation(seed.evaluationId, user)
+                evaluationPage.progressionStepLink(createActionFrom).click()
+            })
+        })
+
+        it('Any nominated user can create an action', () => {
```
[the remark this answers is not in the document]
achaikou: As per agreement, these checks now should go into different test suite.
Thus renamed tests for passive voice to show that the way action performed is what that matters here,

frontend/cypress/integration/actions_spec.ts
achaikou: This fits perfectly for Petter's participants changes, so not bothering with this now.

achaikou (commented):
/azp run

PR #553, opened 2021-08-19
Update Role privileges at Nomination step

ErlendHaa:
This PR solves several issues with who and when different actions could be performed in the Nomination step. It formalises the privilege of Adding new Participants, deleting participants and progressing the evaluation past the Nomination stage:

 - **Add Participant**:  `Facilitator` and `OrganizationLead` can add participants, regardless of which progression the Evaluation is at.
 - **Delete Participants**: `Facilitators` and `OrganizationLead` can delete participants, but only when the Evaluation is still at `Nomination`.
 - **Progress Evaluation**: `Facilitators` can progress evaluation from `Nomination` to `Individual`.
 
 Additional safeguards are added to restrict non-members of the Evaluation from performing any of the above mutations. That was possible prior to this PR.

### DIFF
--- a/frontend/src/utils/RoleBasedAccess.ts
+++ b/frontend/src/utils/RoleBasedAccess.ts
@@ -18,5 +18,7 @@ export const participantCanReadAnswer = (participant: Participant, answer: Answe
                 answer.progression !== Progression.Individual ||
                 participant.id === answer.answeredBy?.id
             )
+        default:
+            return false
     }
 }
--- a/frontend/src/utils/RoleBasedAccess.ts
+++ b/frontend/src/utils/RoleBasedAccess.ts
@@ -22,3 +22,18 @@ export const participantCanReadAnswer = (participant: Participant, answer: Answe
             return false
     }
 }
+
+
+/* Role-based rule for deleting another Participant from the Evaluation */
+export const participantCanDeleteParticipant = (participantRole: Role) => {
+    switch (participantRole) {
+        case Role.Facilitator: // Intentional fall-through
+        case Role.OrganizationLead:
+            return true
+        case Role.Participant: // Intentional fall-through
+        case Role.ReadOnly:
+            return false
+        default:
+            return false
+    }
+}
--- a/frontend/src/utils/RoleBasedAccess.ts
+++ b/frontend/src/utils/RoleBasedAccess.ts
@@ -37,3 +37,18 @@ export const participantCanDeleteParticipant = (participantRole: Role) => {
             return false
     }
 }
+
+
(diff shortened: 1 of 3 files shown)


frontend/src/utils/RoleBasedAccess.ts
```diff
@@ -18,5 +18,52 @@ export const participantCanReadAnswer = (participant: Participant, answer: Answe
...
+        case Role.ReadOnly:
+            return false
+        default:
+            return false
+    }
+}
+
+
+/* Role-based rule for deleting another Participant from the Evaluation */
+export const participantCanDeleteParticipant = (participantRole: Role) => {
```
petterwildhagen, line 43: This does the same as participantCanAddParticipant - consider creating method participantCanAddOrDeleteParticipant ?

frontend/src/views/Evaluation/Nomination/NominationView.tsx
```diff
@@ -4,20 +4,62 @@ import { Box } from '@material-ui/core'
...
+
+/** Who (Role) can add add users, and when (Progression)
+ *
+ * Who:
+ *  Facilitators and OrganizationLeads
+ *
+ * When:
+ *  Any Progression
+ */
+const disableAddParticipant = (evaluation: Evaluation, azureUniqueId: string) => {
```
petterwildhagen, line 48: Consider creating one method disableAddDeleteParticipant. This method does the same as disableDelete

petterwildhagen (approved):
Just one small comment. Otherwise OK

frontend/src/utils/RoleBasedAccess.ts
ErlendHaa: This duplication was intentional. My reasoning was that they are the same now, yes, but these restrictions keep changes it seams. I opt for keeping it as 2 separate function, but internally one can use the other, too save some lines

petterwildhagen: Ok, then this approach is better suited if you expect changes to this,

frontend/src/utils/RoleBasedAccess.ts
```diff
@@ -24,6 +24,13 @@ export const participantCanReadAnswer = (participant: Participant, answer: Answe
 }
 
 
+/* Role-based rule for adding another Participant to the Evaluation */
+export const participantCanAddParticipant = (participantRole: Role) => {
+    /* Currently this share the same access as for deleting Participants */
+    return participantCanDeleteParticipant(participantRole)
+}
```
petterwildhagen, line 31: Nicely solved!!!!

PR #555, opened 2021-08-19
Cypress fixes for contexts call

achaikou:
- Fixes actual contexts situation
- Completely disabled the reason that caused the failure. With that there will be no clearly visible error information for us anymore that we are missing some data, so it has to be provided only when actual functionality is tested. This change is also expected to cause better tests stability.


frontend/cypress/support/mock/external/projects.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+ * Projects below may be considered bad judging by provided properties, but can
+ * be considered ok as of now as we don't truly know what these properties mean.
+ * While `isInvalid`, `isActive`, `isDeleted` are actual returned fields, only
+ * `isActive` is available through `Context` interface.
+ *
+ * At the moment it seems like no invalid, inactive or deleted projects are
+ * present in the real world.
+ */
```
petterwildhagen, line 54: If this cannot appear in the real world, why create test data for it?

petterwildhagen (changes requested):

frontend/cypress/support/mock/external/projects.ts
achaikou: It doesn't mean that it can not. It just means that by analyzing _current_ existing fusion data I found none right now.
But the fields themselves [are](https://github.com/equinor/fusion-functions/blob/master/domains/Fusion.Functions.Context.ContextSync/Models/ProjectFusionContext.cs) [real](https://github.com/equinor/fusion-functions/blob/master/domains/Fusion.Functions.Context.ContextSync/Models/FusionContext.cs). Someone put them there with intention.

Until we get new information, these can be used just as other normal projects (if 3 main ones are not enough for some reason).
I just believed these fields to be interesting enough to create mock objects with them. The same way I created different users for different employee types, thus in theory different privileges.

Meaning that if these interesting fields ever cause any difference in future, we would have a chance to catch it and adjust application logic.

petterwildhagen: If the fields are not used isn't it better to remove them from the application altogether, rather than create testdata for them? It might be 'a bug' that they are there.

achaikou: This is not our code. We have no idea what all the fusion guys plan to do and why they introduced them.

petterwildhagen: The fields are not used in the application. Here is the model for project from frontend/api : export type Project = {
  __typename?: 'Project';
  id: Scalars['String'];
  fusionProjectId: Scalars['String'];
  createDate: Scalars['DateTime'];
  evaluations: Array<Maybe<Evaluation>>;
};  The fields invalid, inactive or deleted  are not present in the model, hence not used in BMT and therefore no testdata needs to or should be created

achaikou:
/azp run

1. Will handle fixups before merge on rebasing. Due to refactoring they do not align correctly to respective comments.
2. Was able to reproduce both errors that are now handled in `on uncaught exception` and verify that they no longer cause problems with current handling
3. Remaining changes done as per discussion.

petterwildhagen (approved):

PR #557, opened 2021-08-19
Do not fail cypress test on uncaught contexts promise

petterwildhagen:
…bsites.net/contexts

### DIFF
--- a/frontend/cypress/support/index.ts
+++ b/frontend/cypress/support/index.ts
@@ -18,3 +18,14 @@ import './visit'
 import './external'
 import './gql'
 import './common'
+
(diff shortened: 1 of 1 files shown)


achaikou:
We have that fix in 3 PRs now :D so need to agree which one goes in :) 

Other things, for future:
1. Please don't forget to format files with prettier.
2. If the commit title doesn't fit into github name, it is in my opinion definitely too long. Too long git commit messages make it difficult to read the log. ([Here is the link](https://chris.beams.io/posts/git-commit/) that goes around in the Equinor chats as a git commit message recommendation, btw)

petterwildhagen:
Close this PR as it is duplicate of other code

PR #558, opened 2021-08-20
Fix backspace behavior in markdown editor

maiaht:


frontend/src/components/QuestionAndAnswer/QuestionAndAnswerForm.tsx
```diff
@@ -56,7 +56,7 @@ const QuestionAndAnswerForm = ({ question, answer, disabled, onAnswerChange, sav
                             </Box>
                             <Box width="85%">
                                 <AnswerMarkdownForm
-                                    markdown={answer.text}
+                                    markdown={answer.text === '' ? ' ' : answer.text}
```
ErlendHaa, line 59: This is the sort of place where we would want an inline comment explaining the code. Without context (i.e. that it fixes a backspace error) the why's of this line is a complete mystery. Hence a inline comment would save us some future wtf's.

maiaht: Fixed!

ErlendHaa (approved):

PR #560, opened 2021-08-20
Create Closing Remarks table in database

maiaht:


backend/api/GQL/Mutation.cs
```diff
@@ -215,6 +218,21 @@ public Note EditNote(string noteId, string text)
             return _noteService.EditNote(note, text);
         }
 
+        public ClosingRemarks CreateClosingRemarks(string actionId, string text)
```
ErlendHaa, line 221: `CreateClosingRemark` - singular

backend/api/Models/Models.cs
```diff
@@ -190,6 +192,21 @@ public class Note
         public virtual Action Action { get; set; }
     }
 
+    public class ClosingRemarks
```
ErlendHaa, line 195: `ClosingRemark` - singular

backend/api/GQL/Mutation.cs
```diff
@@ -215,6 +218,21 @@ public Note EditNote(string noteId, string text)
...
+            Evaluation evaluation = queryableAction.Select(a => a.Question.Evaluation).First();
+
+            return _closingRemarksService.Create(CurrentUser(evaluation), text, action);
+        }
+
+        public ClosingRemarks EditClosingRemarks(string closingRemarksId, string text)
+        {
+            ClosingRemarks closingRemarks = _closingRemarksService.GetClosingRemarks(closingRemarksId);
+            return _closingRemarksService.EditClosingRemarks(closingRemarks, text);
+        }
```
ErlendHaa, line 234: Are we planning on implementing editing of remarks right way? If not, there is no need to introduce this now.
And what about delete?

backend/api/Services/ClosingRemarksService.cs
```diff
@@ -0,0 +1,73 @@
+using System;
+using System.Linq;
+
+using api.Context;
+using api.Models;
+using Action = api.Models.Action;
+
+namespace api.Services
+{
+    public class ClosingRemarksService
```
ErlendHaa, line 10: `ClosingRemarkService`, also applies to the filename

ErlendHaa (commented):
Looks good and works as expected. The naming is a bit off (singular vs plural) and you are missing test for the `ClosingRemarkService`

ErlendHaa (approved):
LGTM!

PR #563, opened 2021-08-23
Move creation of participants out of EvaluationSeed class

petterwildhagen:


frontend/cypress/testdata/progression.ts
```diff
@@ -0,0 +1,5 @@
+import * as faker from 'faker'
+import { Progression } from '../../src/api/models'
+export function getProgression(): Progression {
+    return faker.random.arrayElement(Object.values(Progression))
+}
```
ErlendHaa, line 5: Before I found the implementation, I was pretty confussed as to what this function returned. Maybe `getRandomProgression` or just `randomProgression` is more descriptive?

ErlendHaa (commented):
Looks good! Only nitpicking. The renaming/moving of `users.ts` deserves it's own commit. Mostly to reduce the noise in the current commit.

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -46,15 +47,13 @@ describe('Nomination', () => {
 })
 
 const createDeleteSeed = () => {
+    let users = getUsers(2)
+    let owner = new Participant({ user: users[0], role: Role.Facilitator })
+    let otherFacilitator = new Participant({ user: users[1], role: Role.Facilitator })
     const seed = new EvaluationSeed({
         progression: Progression.Nomination,
-        nParticipants: 2,
+        participants: [owner, otherFacilitator],
```
achaikou, line 55: I suspect we in this test now are in danger of creating too much an invalid state (because backend has little to none restrictions)
1) progression for both facilitators is randomized, so they can end up in FollowUp
2) evaluation is in Nomination stage.

Should that even be possible?

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -127,41 +128,25 @@ export class EvaluationSeed {
...
-        /* The first user will always be a facilitator - and the 'creator' of
-         * the Evaluation*/
-        this.participants[0].role = Role.Facilitator
-
         this.progression = progression
+        this.participants = participants
         this.fusionProjectId = fusionProjectId
-        this.name = evaluationName({prefix: namePrefix})
+        this.name = evaluationName({ prefix: namePrefix })
+        this.participants[0].role = Role.Facilitator;
```
achaikou, line 144: Now you force first participant to be a faciliator, even if I carefully crafted my participants myself where none of them are (for example to test some error combination, idk)

achaikou (commented):
I, on the other hand, have complaints :smiling_imp:

frontend/cypress/testdata/participants.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getOrganisation } from './organization'
+import { Progression, Role } from '../../src/api/models'
+import { getProgression } from './progression'
+
+export function createDefaultParticipants(users: User[], progression?: Progression) {
+    let participants: Participant[] = []
+    users.forEach(user => {
+        if (progression === undefined) {
+            progression = getProgression()
+        }
```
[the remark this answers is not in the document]
petterwildhagen: If I move it outside the loop then all participants will get same progression. Inside the loop each participant will get a random progression assigned

frontend/cypress/testdata/progression.ts
petterwildhagen: Fixed

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -127,41 +128,25 @@ export class EvaluationSeed {
     questions: Question[] = []
 
     constructor({
-        progression = Progression.Individual,
-        nParticipants = 1,
-        users,
+        progression,
+        participants,
```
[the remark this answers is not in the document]
petterwildhagen: It is possible to add a participant after the nomination is complete so addParticipants could reflect, but I have removed the method for now as there is no need for it.

frontend/cypress/integration/nominations_spec.ts
petterwildhagen: Fixed. Both facilitators are now in stage nomination

frontend/cypress/testdata/participants.ts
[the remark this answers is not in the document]
ErlendHaa: Aha, you see I didn't get the "randomness" from looking at this code. But you fixed that 👍

frontend/cypress/testdata/participants.ts
```diff
@@ -0,0 +1,19 @@
+import * as faker from 'faker'
+import { Participant } from '../support/mocks'
+import { getUsers, User } from './users'
+import { getOrganisation } from './organization'
+import { Progression, Role } from '../../src/api/models'
+import { getProgression } from './progression'
+
+export function createDefaultParticipants(users: User[], progression?: Progression) {
```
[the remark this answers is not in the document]
ErlendHaa: > get random progression that follows the rules of evaluation progression

Not saying it's wrong in this case necessarily, but this kind of thing that fast leads us down the road where we replicate the application logic in the test _support_ code.

ErlendHaa:
You should also update the docs on evaluation_seed when you change the interface of the constructor

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -12,17 +12,18 @@ import {
...
+    participants: Participant[]
     fusionProjectId?: string
     namePrefix?: string
 } & (
-    | { nParticipants: number, users?: never }
-    | { users: User[], nParticipants?: never }
-    | { nParticipants?: never, users?: never } // default constructor values
+    | { nParticipants: number; users?: never }
+    | { users: User[]; nParticipants?: never }
+    | { nParticipants?: never; users?: never } // default constructor values
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

frontend/cypress/testdata/participants.ts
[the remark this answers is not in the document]
petterwildhagen: The file participant.ts now only includes createParticipants which creates participants from an array of users where the first user get the role of facilitator. If progression is passed as argument it is set randomly. 
I do not think that the test automation code should enforce business rules - so will leave as is

frontend/cypress/testdata/progression.ts
```diff
@@ -0,0 +1,5 @@
+import * as faker from 'faker'
+import { Progression } from '../../src/api/models'
+export function getProgression(): Progression {
+    return faker.random.arrayElement(Object.values(Progression))
```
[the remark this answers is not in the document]
petterwildhagen: The test designer can assign whatever progression he wants. The testcode should not enforce any business logic.

frontend/cypress/testdata/participants.ts
```diff
@@ -0,0 +1,19 @@
...
+    let participants: Participant[] = []
+    users.forEach(user => {
+        if (progression === undefined) {
+            progression = getProgression()
+        }
+        participants.push(
+            new Participant({ user: user, role: Role.Participant, organization: getOrganisation(), progression: progression })
```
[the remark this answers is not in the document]
petterwildhagen: Added role as optiona, if not specified, then a random role is assgined

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: Fixed by removing

frontend/cypress/support/visit.ts
```diff
@@ -1,4 +1,4 @@
-import { User } from './users'
+import { User } from '../testdata/users'
```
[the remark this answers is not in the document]
petterwildhagen: Nothing wrong with mocks. 
The testdata folder is for code creating testdata. 
I think we should keep mocks folder for code dealing with the mocks, i.e. which URLs to mock.
The testdata folder will contain code that produces testdata.

frontend/cypress/integration/create_evaluation_spec.ts
```diff
@@ -1,45 +1,32 @@
...
 import { EvaluationSeed } from '../support/evaluation_seed'
 import { evaluationName } from '../support/helpers'
 import NominationPage from '../support/nomination'
 import ProjectPage from '../support/project'
-import { User, users } from '../support/mock/external/users'
+import { users, User } from '../support/mock/external/users'
+import * as faker from 'faker'
 
-const evaluationUserIsFacilitator = () => {
+const createEvaluation = (creator: User, otherUser: User, prefix: string) => {
```
ErlendHaa, line 9: Maybe `otherUsers: User[]` to make this helper a bit more helpful. Currently this function can only be used to create evaluations with exactly 2 participants

frontend/cypress/support/mocks.ts
```diff
@@ -26,6 +26,19 @@ export class Participant {
     }
 }
 
+export function createParticipant({ user, role, progression, organization }: IParticipant): Participant {
```
ErlendHaa, line 29: Move this to e.g. `helpers.ts`, it's not a mock

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -12,87 +13,39 @@ import {
...
     CREATE_NOTE,
     SET_SUMMARY,
-    PROGRESS_PARTICIPANT
+    PROGRESS_PARTICIPANT,
 } from './gql'
 
 type EvaluationSeedInput = {
-    progression?: Progression
+    progression: Progression
+    users?: User[]
```
ErlendHaa, line 21: Why would you want this as optional? It's not possible to create such evaluations

frontend/cypress/integration/create_evaluation_spec.ts
petterwildhagen: Yes, that is the intention of this method and it is sufficient for this test that only requires evaluations with two users

frontend/cypress/support/evaluation_seed.ts
achaikou: Well, it is possible to do it now as one can add participants separately and explicitly. So optional fits

petterwildhagen: Agree. Fixed. This also makes EvaluationSeed constructor simpler

frontend/cypress/support/mocks.ts
petterwildhagen: Fixed, moved this and createAction to testdata which improves the organization of the code a lot.

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: In order to create an evaluation you need at least one user, so it makes sense to not have users optional. It is still possible to add participants after creating an evaluation.

ErlendHaa: Alena, it's still possible to create an eval with 0 users as the argument is a list. So removing optional doesn't take away anything, it only makes it explicit

frontend/cypress/support/auth.ts
```diff
@@ -48,6 +48,7 @@ Cypress.Commands.add('login', (user: User = users[0]) => {
             },
         }).then(resp => {
             /* Change to Assertion Function if used more */
+            cy.log('Response ' + resp.headers.location)
```
achaikou, line 51: If you leave this line, at least call it not Response (to general), but "Code flow resolution: auth URL") or something

achaikou (commented):
I don't like the way the development/review process for this PR went. It was quite tiring, also because changes were coming in the middle of the review, which I fiercely hate, and I still can't say everything is good enough here because everything was going so much back and forth for such a small change that I lost track of what is changed and why. Also, PR is tangled without correct split for meaningful commits that make it easier to understand what went down.

I hope with more careful approach there wouldn't be so many review rounds in the future PRs.

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -32,27 +32,28 @@ describe('Nomination', () => {
...
-        users: [],
+        users: users,
     })
-    seed.addParticipant(owner).addParticipant(otherFacilitator)
-    return { seed, owner, otherFacilitator }
+    seed.participants[1].role = Role.Facilitator
+    const facilitator = seed.participants[0]
+    const otherFacilitator = seed.participants[1]
+    seed.addParticipant(facilitator).addParticipant(otherFacilitator)
+    return { seed, facilitator, otherFacilitator }
```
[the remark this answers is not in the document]
petterwildhagen: I missed this when I responded to Erlends comment & you are right, this is incorrect use of the class. Fixed by removing seed.addParticipant(facilitator).addParticipant(otherFacilitator)

frontend/cypress/support/auth.ts
petterwildhagen: Deleted this line.

achaikou (approved):

PR #564, opened 2021-08-24
Add cypress test to Workshop Complete Date

hnformentin:
Solves #526 

Identified four things to think/improve:

(1) Uncaught exceptions (originated from Fusion and not cypress):
- understand what/why it happens (e.g. promises from Fusion?)
- discuss the way we want to treat these exceptions on the code base globally (e.g. create a method/function)

(2) Are methods such as StepperGrid and FollowUpStepperGrid useful for other tests? Otherwise I can have the code directly here.

(3) Reference for tabs changes: 
- Which are the more stable ways to call for objects?
- How to find the reference for elements other that in the navigation option in Cypress interface?

(4) How to change progression status of a seed already planted (so I could use beforeEach, for example)?


frontend/cypress/integration/bownie_completion_date.ts
```diff
@@ -0,0 +1,68 @@
...
+            const unauthorized = /\[https:\/\/pro\-s\-people\-ci\.azurewebsites\.net.+] returned HTTP status code \[401\]/
+            if (err.message.match(unauthorized)) {
+                return false
+            }
+        })
+
+        it('User can not see complete date of uncompleted Evaluation', () => {
+            // need to learn how to change progression after planting the seed
+            // seed.progression = Progression.FollowUp did not work
+            // in this way I could use beforeEach as in actions_spec
```
ErlendHaa, line 27: You can't do that using the seed. A seed is _only_ an initial state. Once a seed is planted (which it should be prior to starting the test) the evaluation is put into the database, and from there on it lives its out it's own life, just like if we had created it using the UI. Once planted, you have to interact with it the same way you would interact with any evaluation. I.e. you have to use the UI to manipulate it. 

So if you want to progress the `Evaluation` from `Workshop` to `Follow-up` as part of your test, you do that by programatically finding the "progress evaluation"  button and pressing it.

frontend/cypress/integration/bownie_completion_date.ts
```diff
@@ -0,0 +1,68 @@
...
+                return false
+            }
+            if (err.message.includes("Cannot read property 'removeEventListener' of null")) {
+                return false
+            }
+            const unauthorized = /\[https:\/\/pro\-s\-people\-ci\.azurewebsites\.net.+] returned HTTP status code \[401\]/
+            if (err.message.match(unauthorized)) {
+                return false
+            }
+        })
```
ErlendHaa, line 22: Yes, and it will be introduced in Alena's PR. It's ok to keep this around here while we wait for her pr. But this should not be merged

frontend/cypress/support/stepper_grid.ts
```diff
@@ -0,0 +1,21 @@
+export default class StepperGrid {
+    nomination = () => {
+        return cy.get('.fc--Stepper__stepPane--1GLbb > :nth-child(1)')
```
petterwildhagen, line 3: Introduce data-testid here so you can write
return cy.get(`[data-testid=nomination-step`)
etc....

frontend/cypress/integration/bownie_completion_date.ts
```diff
@@ -0,0 +1,68 @@
...
+
+            // to test stepper grid
+            // this does not work, the table below become cy.get('#tabs-4-tab-2').click()
+            // stepperGrid.nomination().click()
+            // stepperGrid.individualAssessment().click()
+            // stepperGrid.preparation().click()
+            // stepperGrid.workshopAssessment().click()
+
+            stepperGrid.followUp().click()
+            cy.get('#tabs-2-tab-2').click()
```
petterwildhagen, line 46: Create a component for this tab, or use data-testid to access each tab, 
with component:
tabs.workshop-summary().click()

petterwildhagen (changes requested):

frontend/cypress/support/stepper_grid.ts
ErlendHaa: Note sure that's in our power to do. If I'm not mistaken this is a fusion component

frontend/cypress/integration/bownie_completion_date.ts
```diff
@@ -0,0 +1,68 @@
...
+            // need to learn how to change progression after planting the seed
+            // seed.progression = Progression.FollowUp did not work
+            // in this way I could use beforeEach as in actions_spec
+            const seed = new EvaluationSeed({
+                progression: Progression.Workshop,
+            })
+
+            seed.plant().then(() => {
+                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
+            })
```
ErlendHaa, line 34: This should happen outside the testcase, typically in a `beforeEach`

frontend/cypress/support/stepper_grid.ts
achaikou: Yep. I have all that in the Actions PR and Stepper stuff is really troublesome to identify precisely as, as Erlend says, we can't inject data-testid because it is a fusion component. The best option I found was to identify steps just by text globally. At the moment it doesn't interfere with anything else.

(just one comment, didn't take a look at the whole PR yet)

petterwildhagen: This works: return cy.get('div').contains('Follow-up') 
I tried it out on Helenas branch

achaikou: yep, should work. virtually the same as this
https://github.com/equinor/fusion-bmt/pull/532/files#diff-e4cec85fb9f36f6a626e86a0306679066e3cba3bbae6b8c4b39ecb287fc457f9R34
Both identify components globally, only one depends on "div" structure, other on "a" structure.

frontend/cypress/integration/bownie_completion_date.ts
petterwildhagen: or simply write cy.get('button').contains('Workshop Summary').click() (I tried this out and it works)
A component for this is a bit excessive

frontend/cypress/integration/bownie_completion_date.ts
hnformentin:
```
        beforeEach(() => {
                progression: Progression.Workshop,
            })
            seed.plant().then(() => {
                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
            })
        })
```

followed by:

```
            cy.get('.MuiBox-root-25 > .sc-dlnjwi').click()
            cy.get('[data-testid=yes_button]').click()

```

hnformentin:
Topics in work:
(6) Set/fetch the complete date in a way that we can avoid flake tests (e.g. check that the date is in fact contained in the page)

(7) Investigate the need of `force: true` when clicking some elements (errors are commented in the PR itself)

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,71 @@
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { Progression } from '../../src/api/models'
+import StepperGrid from '../support/stepper_grid'
+import FollowUpTabs from '../support/followup_tabs'
+import FinishButtons from '../support/finish_buttons'
+
+describe('Workshop Summary', () => {
+    context('Bowtie model', () => {
+        // should NOT be merged: Alena's PR code. Should we have a function for that?
+        Cypress.on('uncaught:exception', (err, runnable, promise) => {
```
petterwildhagen, line 10: This can be removed. Otherwise OK

petterwildhagen (approved):

hnformentin:
Even with the great suggestions so far, I am not being able to find alternative ways to `cy.get('div')`, `cy.get('a')` or `cy.get('button')`. I tried to alternate commands with should statements, not successfully until now 👎 
I will get though the new content of `actions_spec.ts` that Alena suggested, maybe there is an alternative already there...

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -59,12 +59,12 @@ describe('Workshop Summary', () => {
 
             finishButtons.workshopAssessment().click({ force: true })
 
-            cy.get('[data-testid=yes_button]').click()
+            cy.get('[data-testid=yes_button]').click({ force: true })
```
ErlendHaa, line 62: Do you need force?

hnformentin: It failed without it now, so I added. But it seems that this PR will need some refactoring anyways, and my aim is to remove the `force` options 👍

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,59 @@
...
+            cy.contains('Workshop completed:').should('not.exist')
+        })
+
+        it('User can see complete date on a complete Evaluation', () => {
+            const today = new Date().toLocaleDateString()
+            const stepperGrid = new StepperGrid()
+            const followUpTabs = new FollowUpTabs()
+            const finishButtons = new FinishButtons()
+
+            finishButtons.workshopAssessment().click({ force: true })
```
hnformentin, line 48: if I try to remove `force:true` in here, locally, I get an error:
Screenshot 2021-08-31 at 10 01 36

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,59 @@
...
+
+            cy.get('[data-testid=yes_button]').click()
+
+            // add documentation/readme use tag name and text to identify elements
+            stepperGrid.followUp().click()
+
+            followUpTabs.workshopSummary().click({ force: true })
```
hnformentin, line 55: And here, without `force:true`, locally:
Screenshot 2021-08-31 at 10 02 56

frontend/cypress/integration/bowtie_completion_date.ts
hnformentin: fixed

hnformentin:
- When reliable selectors are not available, alternating commands and assertions helped to gain stability in the tests
- `force: true` is still required for a couple of buttons (_see_ error messages). It is probably not a best practice, but I did not find alternatives so far
- After review, I will rebase the commits into a single one before merging

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,59 @@
...
+                users: getUsers(2),
+            })
+            seed.plant().then(() => {
+                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
+            })
+        })
+
+        it('User can advance in the stepper grid', () => {
+            // to test stepper grid, maybe should be removed before merging
+            const stepperGrid = new StepperGrid()
```
petterwildhagen, line 24: const stepperGrid = new StepperGrid() can be moved up to below let seed: EvaluationSeed because it is shared between the tests

frontend/cypress/support/stepper_grid.ts
```diff
@@ -0,0 +1,21 @@
+export default class StepperGrid {
+    nomination = () => {
+        return cy.get('a').should('contain', 'Nomination').contains('Nomination')
```
petterwildhagen, line 3: should is an assertion, get and contains retrieves elements.
Retrieve to original  return cy.get('a').contains('Nomination')

hnformentin: Yes, exactly! That was what was causing the instability in Azure pipeline (Alena described above, also https://docs.cypress.io/guides/core-concepts/retry-ability#Only-the-last-command-is-retried).
Basically, iIf loading the elements is too slow and `Nomination` is not found after 4s, the test fails because only the last command is tried. 

One of the solutions mentioned in Cypress documentation is to alternate commands and assertions as implemented now in the PR.
The assertions in-between commands make sure that `Nomination` is among the `a`s found, if not it keeps retrying this assertion command before proceeding to `contains`.

hnformentin: In this way the test does not fail in the pipeline as before...

achaikou: (didn't review)

1. We still need to combine existing code and code you introduced. Now we have stepper in two files - `stepper_grid.ts` and `evaluation.ts` (`progressionStepLink`).  Please leave just one method to live?
2. Duplication might apply to some other methods as well. See the current logic between how support files names and locations are chosen - we need to make sure that we are always able to find the locators other people already introduced. Now in merged code we go with pages/views/features - something not too small, but not too big for size of our app. This solution still can be changed in the future, but for now it exists, so maybe new code must be compliant to it.
Like for example, Finish buttons are not a separate thing. They definitely are common for evaluation in any stage, so probably should be present in the `evaluation.ts` instead of in a separate file. followup_tabs.ts maybe better be renamed to `followup.ts` for now - to keep the file most general. We don't want too many support files (at least at the moment), so having just 1 file for 4 follow up tabs seems a bit too much. Keeping them in the whole followup component seems to make more sense as everything that belongs to the followup page can exist there (unless there is a big or duplicated between pages feature that can be moved to a separate file)
3. While I agree with you that this solution is valid and should work fine, it does look a bit weird. Maybe you can use just the simple form and add [selector](https://docs.cypress.io/api/commands/contains#Arguments) to `contains` aka
`return cy.contains('a', "Nomination")`
?

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,59 @@
...
+import StepperGrid from '../support/stepper_grid'
+import FollowUpTabs from '../support/followup_tabs'
+import FinishButtons from '../support/finish_buttons'
+import { getUsers } from '../support/mock/external/users'
+
+describe('Workshop Summary', () => {
+    context('Bowtie model', () => {
+        let seed: EvaluationSeed
+
+        beforeEach(() => {
```
petterwildhagen, line 12: Replace with before(). You can reuse the same state between tests (it('...)). This will also speed up the test

hnformentin: Fixed together, thanks!

frontend/cypress/support/stepper_grid.ts
hnformentin: 1. `progressionStepLink` substituted `StepperGrid`
2. Moved `FinishButtons` to `evaluation.ts`; `followup_tabs.ts` renamed to `followup.ts`, keeping `FollowUpTabs` as the class name. Does it seem ok?
3. As StepperGrid was removed in "1" this was not necessary, but changed in `cy.contains('#fixed-tablist', 'Workshop Summary')`

frontend/cypress/integration/bowtie_completion_date.ts
hnformentin: Good point! Fixed.

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        it('User can see complete date on a complete Evaluation', () => {
+            const today = new Date().toLocaleDateString()
+            const finishButtons = new FinishButtons()
+            const confirmationDialog = new ConfirmationDialog()
+
+            cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
+
+            evaluationPage.progressionStepLink(Progression.Workshop).click()
+            finishButtons.workshopAssessment().click()
```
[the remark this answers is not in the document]
hnformentin: I understand your point. I would then initiate the seed directly on a random state except `Follow-up` (and not have the date) or in the `Follow-up` state (and have the date displayed).  Would that be in the direction of what you described about the use of `seed.plant().then(() => {})`?

`
import { EvaluationSeed } from '../support/evaluation_seed'
import { Progression } from '../../src/api/models'
import FollowUpTabs from '../support/followup'
import { EvaluationPage } from '../support/evaluation'
import { getUsers } from '../support/mock/external/users'

describe('Workshop Summary', () => {
    let seed: EvaluationSeed
    const evaluationPage = new EvaluationPage()
    const followUpTabs = new FollowUpTabs()

    context('Bowtie model', () => {
        it('User can not see complete date on an incomplete Evaluation', () => {
            seed = new EvaluationSeed({
                progression: Progression.Workshop,
                users: getUsers(2),
            })
            seed.plant().then(() => {
                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
                evaluationPage.progressionStepLink(Progression.FollowUp).click()
                followUpTabs.workshopSummary().click()

                cy.contains('Workshop completed:').should('not.exist')
            })
        })

        it('User can see complete date on a complete Evaluation', () => {
            const today = new Date().toLocaleDateString()

            seed = new EvaluationSeed({
                progression: Progression.FollowUp,
                users: getUsers(2),
            })
            seed.plant().then(() => {
                cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
                followUpTabs.workshopSummary().click()

                cy.contains(`Workshop completed: ${today}`).should('exist')
            })
        })
    })
})
`

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,47 @@
...
+import { FinishButtons, EvaluationPage } from '../support/evaluation'
+import { getUsers } from '../support/mock/external/users'
+import { ConfirmationDialog } from '../support/common'
+
+describe('Workshop Summary', () => {
+    let seed: EvaluationSeed
+    const evaluationPage = new EvaluationPage()
+    const followUpTabs = new FollowUpTabs()
+
+    before(() => {
```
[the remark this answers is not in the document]
hnformentin: For me both arguments are valid 👍 : (a) Do not repeat yourself, (b) Design independent tests. 
The point you made in the previous comment would answer both arguments, so I think we could stick with it, avoiding both `before` and `beforeEach`.

frontend/cypress/integration/bowtie_completion_date.ts
```diff
@@ -0,0 +1,47 @@
...
+        seed = new EvaluationSeed({
+            progression: Progression.Workshop,
+            users: getUsers(2),
+        })
+        seed.plant()
+    })
+
+    context('Bowtie model', () => {
+        it('User can not see complete date on an incomplete Evaluation', () => {
+            cy.visitEvaluation(seed.evaluationId, seed.participants[0].user)
```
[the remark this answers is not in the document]
hnformentin: Randomizing is another good point. I will implement it.

frontend/cypress/support/followup.ts
```diff
@@ -0,0 +1,5 @@
+export default class FollowUpTabs {
+    workshopSummary = () => {
+        return cy.contains('#fixed-tablist', 'Workshop Summary')
```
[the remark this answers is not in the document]
hnformentin: Good point. I indeed misunderstood. Will implement your suggestion!

achaikou (approved):
Looks good to me! :+1: 
I still believe that "be.visible" instead of "exists" would make more sense here, given that you explicitly go to a different tab, and that we got burned by it already, but it's not really important.
When rebasing, take a look at the Codacy warnings, they make sense this time. They help us who are not familiar enough with typescript to use language better.

hnformentin:
Thanks @achaikou!
- Only now I understand what you meant by "be.visible" instead of "exists", so I made the change for the second test. Though I thought for the first one we could keep "not.exist" in any case.
- One of the suggestions of Codacy was to shorten "progression: progression," to "progression," or renaming "progression". After chatting with @ErlendHaa , my conclusion was that it would be better to rename "progression" to something like "randomProgressionExceptFollowup". As I had already forced push once and run the cypress pipeline, I opt for not to changing this, not a big difference at the end. 
- The failure in Azure pipeline was on the other tests (actions_spec.ts and create_evaluation_spec.ts), what seems to be the expected errors until the other PR is merged.
- After standup I will merge.

PR #566, opened 2021-08-24
Fixed bug where new note was not listed

denektenina:
**The bug**: In the action list on either the Dashboard (1), or in the Follow-up tab (2),

(1)
image

(2)
image

when clicking on an action to open it in the side panel, and then writing a note, the note will not appear in the list of notes beneath the input field, even though the saved icon appears and the note seems to be saved. This problem does not happen when opening an action in the side panel in the questions view (3)

(3)
image

**This fix:** The problem was that we were sending a copy of the action to the side panel, and this copy had the old list of notes which then would not list the new note. Now, the action in the side panel is refreshed every time the component is refreshed, and will thus have the newest data.

### DIFF
--- a/frontend/src/components/ActionTable/ActionTableForOneUserWithApi.tsx
+++ b/frontend/src/components/ActionTable/ActionTableForOneUserWithApi.tsx
@@ -65,8 +65,8 @@ const ActionTableForOneUserWithApi = ({ azureUniqueId }: Props) => {
     const [projects, setProjects] = useState<Context[]>()
     const [isFetchingProject, setIsFetchingProject] = useState<boolean>(true)
     const [isEditSidebarOpen, setIsEditSidebarOpen] = useState<boolean>(false)
-    const [actionToEditId, setActionToEditId] = useState<string>('')
-    const actionToEdit = actionsWithAdditionalInfo.find(actionWithInfo => actionWithInfo.action.id === actionToEditId)
+    const [actionIdToEdit, setActionIdToEdit] = useState<string>('')
+    const actionToEdit = actionsWithAdditionalInfo.find(actionWithInfo => actionWithInfo.action.id === actionIdToEdit)
 
     useEffect(() => {
         apiClients.context.getContextsAsync().then(projects => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/ActionTable/ActionTableForOneUserWithApi.tsx
```diff
@@ -65,7 +65,8 @@ const ActionTableForOneUserWithApi = ({ azureUniqueId }: Props) => {
     const [projects, setProjects] = useState<Context[]>()
     const [isFetchingProject, setIsFetchingProject] = useState<boolean>(true)
     const [isEditSidebarOpen, setIsEditSidebarOpen] = useState<boolean>(false)
-    const [actionToEdit, setActionToEdit] = useState<ActionWithAdditionalInfo | undefined>()
+    const [actionToEditId, setActionToEditId] = useState<string>('')
```
maiaht, line 68: Could this variable be called `actionIdToEdit` instead of `actionToEditId`?

frontend/src/components/ActionTable/ActionTableWithApi.tsx
```diff
@@ -39,17 +39,17 @@ const ActionTableWithApi = ({ evaluations }: Props) => {
 
     const { personDetailsList } = useAllPersonDetailsAsync(allAzureUniqueIds)
     const [isEditSidebarOpen, setIsEditSidebarOpen] = useState<boolean>(false)
-    const [aqeToEdit, setAqeToEdit] = useState<ActionQuestionAndEvaluation | undefined>()
+    const [actionToEditId, setActionToEditId] = useState<string>('')
```
maiaht, line 42: `actionToEditId` -> `actionIdToEdit` here as well?

frontend/src/components/ActionTable/ActionTableForOneUserWithApi.tsx
denektenina: It actually was called that earlier, but since this is the Id now, and not the actual action, I felt it was maybe clearer to give it a more descriptive name so that it doesn't look like it can be sent directly to the Sidepanel like earlier (now we have to find the right action using the id)

frontend/src/components/ActionTable/ActionTableWithApi.tsx
denektenina: Same answer as above :)

frontend/src/components/ActionTable/ActionTableForOneUserWithApi.tsx
denektenina: After clarification on slack I see id is still in there. Can change that, I agree it's better naming

maiaht (approved):

PR #579, opened 2021-08-31
Cypress Nomination tests

petterwildhagen:


frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+                addUser: true,
+                deleteUser: false,
+                progressEval: false,
+            },
+            {
+                role: Role.Participant,
+                addUser: false,
+                deleteUser: false,
+                progressEval: false,
+            },
```
ErlendHaa: We should add a case for `Role.Read-only` too

ErlendHaa (commented):
The last commit contains a little bit more than you intended I think

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -132,6 +133,17 @@ export class EvaluationSeed {
         return question.id
     }
 
+    findRandomParticipant(role: Role): Participant {
```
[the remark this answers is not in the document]
petterwildhagen: Moved it to testdata analogous to createAction, createParticipant

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -132,6 +133,17 @@ export class EvaluationSeed {
...
+        let participants: Participant[] = []
+        this.participants.forEach(x => {
```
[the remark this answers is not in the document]
petterwildhagen: Fixed. Replaced by filter

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+    seed.participants[3].role = Role.OrganizationLead
+    return seed
+}
+
+interface IVerifyUserManagementCapabilities {
+    participant: Participant
+    addUser: boolean
+    deleteUser: boolean
+    progressEval: boolean
+}
```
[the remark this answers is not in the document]
petterwildhagen: Fixed. Removed interface. This compresses the code so it is better

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+    if (progressEval) {
+        nominationPage.finishNominationButton().should('be.enabled')
+    } else {
+        nominationPage.finishNominationButton().should('be.disabled')
+    }
+    // Delete button never visible for user itself regardless of role
+    nominationPage.deletePersonDiv(participant.user).should('not.exist')
+    allParticipants.forEach(p => {
+        if (p === participant) {
+            return
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+                deleteUser: true,
+                progressEval: false,
+            },
+            {
+                role: Role.Participant,
+                addUser: false,
+                deleteUser: false,
+                progressEval: false,
+            },
+        ].forEach(e => {
```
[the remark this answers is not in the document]
petterwildhagen: Fixed, but left forEach instead of for loop

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+                role: Role.Facilitator,
+                addUser: true,
+                deleteUser: true,
+                progressEval: true,
+            },
+            {
+                role: Role.OrganizationLead,
+                addUser: true,
+                deleteUser: true,
+                progressEval: false,
```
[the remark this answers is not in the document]
petterwildhagen: Fixed.

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -132,6 +133,7 @@ export class EvaluationSeed {
         return question.id
     }
 
+    public findRandomParticipant = findRandomParticipant
```
ErlendHaa, line 136: I fail to see the point in re-defining it here

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,164 @@ import { Progression, Role } from '../../src/api/models'
...
+describe('Nomination stage', () => {
+    describe('Adding and deleting users, finishing nomination', () => {
         let seed: EvaluationSeed
-        let facilitator: Participant
-        let otherFacilitator: Participant
-
-        beforeEach(() => {
-            ;({ seed, facilitator, otherFacilitator } = createDeleteSeed())
+        const nominationPage = new NominationPage()
+        before(() => {
```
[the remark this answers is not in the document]
petterwildhagen: To be able to write 
let p = seed.findRandomParticipant(e.role) , that is get a random participant directly from the seed object.  If I define it outside the class, it reads 
let p = findRandomParticipant(seed, e.role)
I just think the former is more readable.

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: To be able to write
let p = seed.findRandomParticipant(e.role) , that is get a random participant directly from the seed object. If I define it outside the class, it reads
let p = findRandomParticipant(seed, e.role)
I just think the former is more readable.

ErlendHaa: Maybe. My concern is that in the process of a creating a slightly more readable function you are also extending the definition of the seed class. I predict that in the future we will have a high number of similar helper/convenience-functions that all can be added to this class with the same rationale. And that will obscure it's original purpose

frontend/cypress/integration/nominations_spec.ts
petterwildhagen: Done. I overlooked this first.

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: I agree there  will be quite a few of these functions because we need the ability to query EvaluationSeed (i.e. what actions are there etc). The original purpose being to have class as a specification for testdata to be set for evaluation. We will also need the ability to query  EvaluationSeed. Either these functions reside in the tests (close to where they are used) or in the class. In the latter case it might get messy - so I can move it to the test itself.

frontend/cypress/support/stepper_grid.ts
```diff
@@ -0,0 +1,21 @@
+export default class StepperGrid {
```
achaikou, line 1: (didn't review, just one warning)

You might get bitten by same problems as Helena.
1) this functionality is already merged in `evaluation.ts` so lets not duplicate it
2) avoid using `cy.get('a').contains("something")` and use `cy.contains('a', "something")` instead or you, depending on surrounding app state, might get stability surprises on CI https://docs.cypress.io/guides/core-concepts/retry-ability#Only-the-last-command-is-retried

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: Fixed

frontend/cypress/support/stepper_grid.ts
petterwildhagen: I just copied this one from Helenas PR because I need it. Is it this? progressionStepLink = (progression: Progression) ?
I did not guess because it is in evaluations.ts ... But I can use that instead and not use the stepper. We should consider renaming from progressionStepLink to progressionStepper (it is not really a link)

petterwildhagen: Fixed, stepper_grid replaced with progressionStepLink. The naming should reflect what the user thinks it is, not the technical implemantation. I think progressionStepper is closer. A link suggests that you are taken somewhere else, which is not the case. The application also does not really have pages, but it is more like it has stages. So at some point I think we should adjust our naming a bit.

frontend/cypress/integration/nominations_spec.ts
```diff
@@ -3,56 +3,181 @@ import { Progression, Role } from '../../src/api/models'
...
+        progression: progression,
+        users,
     })
     seed.participants[1].role = Role.Facilitator
-    const facilitator = seed.participants[0]
-    const otherFacilitator = seed.participants[1]
-    return { seed, facilitator, otherFacilitator }
+    seed.participants[2].role = Role.Participant
+    seed.participants[3].role = Role.OrganizationLead
+    seed.participants[4].role = Role.ReadOnly
```
ErlendHaa, line 141: Also add a test where you log in as a non-participant and check the same permissions? Sorry forgot about that the last time around

petterwildhagen: The test already does this, see lines 46-47 & 121-122.

ErlendHaa: Sorry my bad

ErlendHaa (approved):
The git log contains a lot of intermediate commits that are irrelevant in the long run. Approved, but rebase before merge

PR #582, opened 2021-08-31
Explain how to work with migrations in README

ErlendHaa:
This commits attempts to clarify the workflow of creating and applying
new database migrations using EF Core.

Closes #570

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,11 +70,6 @@ If you wish to interact with endpoints directly during development, [disable aut
 
 The Schema used for the models in the backend can be found [here](https://backend-fusion-bmt-dev.radix.equinor.com/graphql?sdl).
 
-### Update frontend schema
-
-When the data models in the backend changes, the corresponding models in frontend has to be updated as well. This is done by executing the command `npm run schema` in the fronend directory while the backend server is also running. Then there will be created an updated `models.ts` in the src/api folder which contains the updated models.
-Note that to run this command you also might need to [disable authorization](#disable-authorization).
-
 ### Disable authorization
 
 It is sometimes useful to disable authorization for local development.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+or simply delete the files and changes created by `add`. Once deleted you can
+make new changes to the model and then create a new migration with `add`.
+
+### Updating the frontend schema
+
+Our frontend also needs to know about the current database schema. Once a new
+migration is created, run:
+
+```bash
+npm run schema
```
achaikou, line 148: Now from the docs it looks like migrations are the only case where you want to run schema, which is not true.
When you add some new mutations or queries without changing the db state, you still need to do so.

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+You can apply migrations to the test database at any time to test that it
+behaves as expected.
 
-Make sure you have dotnet-ef installed: `dotnet tool install --global dotnet-ef`
-and that you have set your `Database__ConnectionString`.
+Once the changes are merged to master you should apply the migration to the dev
+database.
 
-Check migrations existing in DB: `dotnet ef migrations list`
+The prod database doesn't need to updated manually, as all migrations are
```
achaikou, line 186: need to (be) updated

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+```
+
+`add` will make changes to existing files and add 2 new files in
+`backend/api/Migrations`, which all needs to be checked in to git.
+
+Note that the {migration name} is just a descriptive name of your choosing.
+Also note that `Database__ConnectionString` should be pointed at one of our
+databases when running `add`.
+
+If you for some reason is unhappy with your migration, you can delete it with
```
achaikou, line 133: If you .... (are) unhappy

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+For the migration to take effect, we need to apply it to our databases. To get
+an overview of the current migrations in a database, set the correct
+`Database__ConnectionString` for that database and run:
+
+```bash
+dotnet ef migrations list
+```
+
+This will list all migrations that are applied to the database and the local
+migrations that are yet to be applied. The latter is denoted with the text
```
achaikou, line 166: "The latter" aka the local migrations, so probably:
The latter are denoted

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+(pending).
+
+To apply the pending migration to the database run:
```
achaikou, line 169: pending migration(s)?
I am not sure, but assume that update will apply all migrations at once? Or would I have to update for every pending migration?

achaikou (commented):
:tulip: to the docs master!
Can't check the flow correctness, as never did any migration, but seems reasonable. Maybe someone else can take a look.

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+Our frontend also needs to know about the current database schema. Once a new
+migration is created, run:
+
+```bash
+npm run schema
+```
+
+from the `/frontend`-directory. This will update the frontend schema. These
+changes must also be checked in to git. Note that for `npm run schema` to run
+properly the backend must be running and authentication must be turned off.
```
maiaht, line 153: Maybe there could be a link here to the "Disable authorization" section in the readme?

README.md
```diff
@@ -104,20 +99,94 @@ Cypress tests will be run in Azure DevOps when pushing to the upstream branch cy
...
+After making changes to the model run (from `/backend/api`):
+
+```bash
+dotnet ef migrations add {migration-name}
+```
+
+`add` will make changes to existing files and add 2 new files in
+`backend/api/Migrations`, which all needs to be checked in to git.
+
+Note that the {migration name} is just a descriptive name of your choosing.
```
maiaht, line 129: migration name -> migration-name

ReedOnly (approved):

README.md
```diff
@@ -31,6 +31,21 @@ npm install
 npm start
 ```
 
+### Updating the frontend schema
+
+Any changes made to the database [model](backend/api/Startup.cs) or the GraphQL
+[queries](backend/api/GQL/Query.cs) and
+[mutations](backend/api/GQL/Mutation.cs) in the backend needs to be
```
achaikou, line 38: need()

as changes... need to be

README.md
```diff
@@ -89,41 +86,106 @@ cd backend/api
...
+dotnet ef migrations add {migration-name}
+```
 
-Make sure you have dotnet-ef installed: `dotnet tool install --global dotnet-ef`
-and that you have set your `Database__ConnectionString`.
+`add` will make changes to existing files and add 2 new files in
+`backend/api/Migrations`, which all need to be checked in to git.
+
+Note that the {migration-name} is just a descriptive name of your choosing.
+Also note that `Database__ConnectionString` should be pointed at one of our
```
achaikou, line 117: Still think we need to explicitly mention not "any of dbs", but "most up to date db" at least, aka dev one, probably... but ok

achaikou (approved):

PR #586, opened 2021-09-01
Revise actions test

petterwildhagen:
Initial state in database is set in beforeEach for each test.
The the test data used in the test is retrieved or created in the test itself.

PR #593, opened 2021-09-03
Cypress test edit action in dashboard actions list

denektenina:


frontend/cypress/integration/actions_spec.ts
```diff
@@ -96,25 +97,77 @@ describe('Actions', () => {
...
+        let updatedAction: Action
+        let newNotes: Note[]
+
+        beforeEach(() => {
+            seed = new EvaluationSeed({
+                progression: faker.random.arrayElement(Object.values(Progression)),
+                users: getUsers(faker.datatype.number({ min: 1, max: 4 })),
             })
+            editor = faker.random.arrayElement(seed.participants)
+            ;({ existingAction, existingNotes } = addActionsAndNotesToEvaluation(seed, {
```
petterwildhagen, line 126: Move into test

frontend/cypress/integration/actions_spec.ts
```diff
@@ -96,25 +97,77 @@ describe('Actions', () => {
...
+                title: 'My test action',
+                assignedTo: editor,
+                completed: false,
+            }))
+            ;({ updatedAction, newNotes } = createEditTestData(seed, editor, existingAction, false))
```
petterwildhagen, line 131: Move testdata creation into test

frontend/cypress/integration/actions_spec.ts
denektenina: Fixed

frontend/cypress/integration/actions_spec.ts
denektenina: Fixed

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+                editActionDialog.assertActionValues(updatedAction, existingNotes.concat([newNote]))
+            })
+        }
+
+        it('Action can be completed without writing a reason', () => {
+            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
+            editActionDialog.assertViewActionNotCompleted()
+            editActionDialog.openCompleteActionView()
+            editActionDialog.confirmAndCheckCompleted()
+            cy.wait(WAIT)
```
hnformentin, line 146: Why is it necessary WAIT here? I tested locally and it passes without it, reading the line 13, I still could not get the reason.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+            editActionDialog.assertActionValues(updatedAction, existingNotes.concat([newNote]))
```
hnformentin, line 147: In `editActionDialog.assertActionValues`, there is a comment `// TODO: time not checked. Need to figure out if it can be done reasonably`.
Do you think it would be reasonable to implement it using something around:

```
            const today = new Date().toLocaleDateString()
            const timeNow = new Date().toLocaleTimeString().substring(0, 3)
             // only hours are checked, since adding minutes could be flaky

            cy.contains(`${today}`).should('be.visible')
            cy.contains(`${timeNow}`).should('be.visible')
```

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+            editActionDialog.close()
+            testCacheInReopenedActionView()
+        })
+
+        it('Action can be completed with a reason', () => {
```
hnformentin, line 152: Maybe 'Action can be completed with writing a reason' is this test or 'Action can be completed without a reason' in the previous one to add consistency.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+            cy.wait(WAIT)
+            editActionDialog.assertActionValues(updatedAction, existingNotes.concat([newNote]))
+            editActionDialog.close()
+            testCacheInReopenedActionView()
+        })
+
+        it('Completing Action can be cancelled', () => {
+            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
+            editActionDialog.assertViewActionNotCompleted()
+            editActionDialog.openCompleteActionView()
```
hnformentin, line 167: This three lines above are the same in the three tests.
What do you think about adding it in beforeEach?

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+        function testCacheInReopenedActionView() {
+            cy.testCacheAndDB(() => {
+                evaluationPage.progressionStepLink(actionProgression).click()
+                actionsGrid.actionLink(existingAction.questionOrder, updatedAction.title).click()
+                editActionDialog.assertActionValues(updatedAction, existingNotes.concat([newNote]))
+            })
+        }
+
+        it('Action can be completed without writing a reason', () => {
+            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
```
hnformentin, line 142: Would not `getActionLink` be a better name here?
My impression from just reading the code is that it would be more consistent with the coming commands (e.g. assert, open, confirm) and also for not forgetting to use/read `click()` in the end...

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+        it('Action can be completed with a reason', () => {
+            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
+            editActionDialog.assertViewActionNotCompleted()
+            editActionDialog.openCompleteActionView()
+            editActionDialog.completedReasonInput().replace('Closed because of a good reason')
+            editActionDialog.confirmAndCheckCompleted()
+            cy.wait(WAIT)
+            editActionDialog.assertActionValues(updatedAction, existingNotes.concat([newNote]))
+            editActionDialog.close()
+            testCacheInReopenedActionView()
```
hnformentin, line 161: Do you see a good way to put the last four lines into a function?
My reasoning is that it would avoid repeating the code into two tests (this and the previous one), which together with adding 
```
            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
            editActionDialog.assertViewActionNotCompleted()
            editActionDialog.openCompleteActionView()
```
to beforeEach, would highlight that the only difference between the two first tests is `editActionDialog.completedReasonInput().replace('Closed because of a good reason')`

petterwildhagen:
In Dashboard Actions table, actions can be edited and completed in side panel
PETTER: Who can edit and complete actions ?
In FollowUp Actions table, actions can be edited and completed in side panel
PETTER: Who can edit and complete actions?
(New) Actions are visible in Dashboard Actions tab
(New) Actions are visible in FollowUp Actions tab

I recently fixed a bug where newly created notes did not appear in the notes list until the side panel were closed and reopened again. The existing tests did not test that, and this issue were partly opened because of this. Therefore, I considered it important to make a test that covered this. This was difficult to test because the notes call was often slow to return, so I landed on setting a wait for 4 seconds before asserting the existence in the notes list. 4 seconds is because cypress automatically fails any calls that take longer than that to return, making it consistent. It makes the tests slower since many of them are waiting longer than they have to, but I think this is important to test.
PETTER: Avoid long waits - they make the tests flaky - investigate why it takes so much time and fix root cause (the slowness)

frontend/cypress/integration/actions_spec.ts
petterwildhagen, line 146: Avoid static waits

frontend/cypress/integration/actions_spec.ts
```diff
@@ -117,6 +109,68 @@ describe('Actions', () => {
...
+        })
+
+        it('Action can be completed with a reason', () => {
+            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
+            editActionDialog.assertViewActionNotCompleted()
+            editActionDialog.openCompleteActionView()
+            editActionDialog.completedReasonInput().replace('Closed because of a good reason')
+            editActionDialog.confirmAndCheckCompleted()
+            cy.wait(WAIT)
```
petterwildhagen, line 158: AVoid WAIT

frontend/cypress/integration/actions_spec.ts
petterwildhagen: I dont this should be in beforeEach. beforeEach should be reserved for setting up the system for test.
The steps above are steps that make up what the test does

hnformentin: Maybe on a function? Or some technique to avoid repeating code three times...

frontend/cypress/integration/actions_spec.ts
denektenina: I wrote the reason in the description and on the constant :)

frontend/cypress/integration/actions_spec.ts
denektenina: Sounds good to me!

frontend/cypress/integration/actions_spec.ts
denektenina: Yes, I can change that

frontend/cypress/integration/actions_spec.ts
denektenina: I agree that a function sounds good!

frontend/cypress/integration/actions_spec.ts
denektenina: Yeah, probably. It's not my naming, so I don't know if there are any other thoughts / conventions behind it

frontend/cypress/integration/actions_spec.ts
denektenina: True!

hnformentin:
My perception into the high level structure of Action tests is that the organisation is not in a structured way. 

For example, I would think that _Actions Tab on Dashboard / Actions can be edited_ would more naturally fit into something like _Edit / Action can be edited from Action Tab on Dashboard _ and so on; _Complete action _ would be more consistent with other contexts if renamed to _Complete_.

I am probably missing something on the way the overall structure was thought, so please share your thoughts :-) 

Screenshot 2021-09-06 at 15 14 08
Screenshot 2021-09-06 at 15 14 57

frontend/cypress/integration/actions_spec.ts
denektenina: Why do the tests become flaky because of long wait? I actually set a long wait to avoid inconsistency (they always get 4 seconds to return, and if the require more it's ok if they fail because cypress doesn't allow more then 4 seconds either). But I think 1 second should be enough, but that's when we risk what I though you ment by "flaky", i.e. they fail sometimes and pass sometimes depending on if they were able to return within a second.

denektenina: Like I wrote in my description, I think this is important to test since we already had a bug on this. If I should remove this wait it would be good if you either argue a better way of doing it, or argue why this isn't so important to test

frontend/cypress/integration/actions_spec.ts
```diff
@@ -240,85 +294,130 @@ describe('Actions', () => {
...
+            editActionDialog.assertActionValues(completedAction, existingNotes.concat([newNote]))
+            editActionDialog.close()
+
+            cy.testCacheAndDB(() => {
+                dashboardActions.tab().click()
+                dashboardActions.actionLink(existingAction.id, existingAction.title).click()
+                editActionDialog.assertActionValues(completedAction, existingNotes.concat([newNote]))
+            })
```
hnformentin, line 362: This is more a question than a suggestion.
The last part of this test is the same as the test above, except the variables `completeAction` and `updatedAction`.
Do you think it would be the case to use `AfterEach` or a function for that?

This would help me to understand a general question: when it is advised to repeat code, when it is better to join it into a function, when AfterEach is the proper structure... 
Thanks to @petterwildhagen comment above I understand now the difference between `beforeEach` (set state before test) and a function (which not necessarily need to return something, as I previously though).

frontend/cypress/integration/actions_spec.ts
```diff
@@ -357,13 +464,13 @@ const createEvaluationWithActionsWithNotes = (actionParameters: Partial<Action>)
         })
     })
 
-    const otherActions: Action[] = Array.from({ length: faker.datatype.number({ min: 0, max: 2 }) }, () => {
+    const otherActions: Action[] = Array.from({ length: faker.datatype.number({ min: 0, max: 5 }) }, () => {
```
hnformentin, line 467: This makes me think about a rhetorical question: why 5 and not 2 or 3, 4 ...?
If our convention is to use `random` when the value is not constraint, shouldn't we set a random number to express when we do not meant a specific corner case (e.g. in this case it is not important if it is two or five actions, if I did not misunderstand).

frontend/src/views/Project/ProjectRoute.tsx
```diff
@@ -43,13 +43,14 @@ const ProjectRoute = ({ match }: RouteComponentProps<Params>) => {
             <Tabs activeTab={activeTab} onChange={setActiveTab}>
                 <TabList>
                     <Tab>Dashboard</Tab>
-                    <Tab>Actions</Tab>
+                    <Tab data-testid="dashboard_actions_tab">Actions</Tab>
```
hnformentin, line 46: Very good that you implemented the `data-testid` here and `id` above 🍾

frontend/cypress/integration/actions_spec.ts
hnformentin: Maybe @ErlendHaa @achaikou know about this one...

frontend/cypress/integration/actions_spec.ts
hnformentin: Sorry if this wait discussion seems repetitive, it is not the intention.
Nevertheless, I did not get your point in the last comment @denektenina .
Did you design the test with wait (a) to reflect a behaviour that you foresee in the production code, (b) as  a way to make sure the elements to be tested were loaded, (c) other option? 😸

frontend/cypress/integration/actions_spec.ts
```diff
@@ -240,85 +294,130 @@ describe('Actions', () => {
...
+            const saveWithComment = Math.random() < 0.5
+            const { updatedAction: completedAction, newNote } = createCompleteActionData(seed, editor, existingAction)
+
+            editActionDialog.assertViewActionNotCompleted()
+            editActionDialog.openCompleteActionView()
+            if (saveWithComment) {
+                editActionDialog.completedReasonInput().replace('Closed because of a good reason')
+            }
+            editActionDialog.confirmAndCheckCompleted()
+            cy.wait(WAIT)
```
petterwildhagen, line 413: Rewrite to avoid wait

frontend/cypress/integration/actions_spec.ts
petterwildhagen: You could parameterize it (one test with text = null, another with text = Closed because of a good reason' for example.

frontend/cypress/integration/actions_spec.ts
petterwildhagen: Yes, we should investigate why is fails. Such failures might point to a problem in the architecture.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -195,14 +249,11 @@ describe('Actions', () => {
...
+        let editor: Participant
 
         beforeEach(() => {
-            ;({ seed, actions } = createViewSeed())
-
-            seed.plant().then(() => {
-                const user = faker.random.arrayElement(seed.participants).user
-                cy.visitEvaluation(seed.evaluationId, user)
-            })
+            ;({ seed, editor, actions } = createViewSeed())
```
petterwildhagen, line 255: Only create seed in beforeEach().
Extract actions in the test itself. Also extract the user in the test. See https://github.com/equinor/fusion-bmt/pull/586 for examples
The word 'editor' does not exist. User other word.

petterwildhagen (changes requested):

frontend/cypress/integration/actions_spec.ts
```diff
@@ -240,85 +294,130 @@ describe('Actions', () => {
...
-        }
-
-        function openConfirmCompleteView() {
-            editActionDialog.completeActionButton().click()
-            editActionDialog.completeActionButton().should('be.disabled')
-            editActionDialog.completedReasonInput().should('exist')
-            editActionDialog.completeActionConfirmButton().should('exist')
-            editActionDialog.completeActionCancelButton().should('exist')
-        }
+        it('Action can be edited', () => {
```
achaikou, line 325: Something important for me: are your newly added tests actually interesting enough to be run every single time?

My original idea was just to add randomization of entry point to the existing complete/edit test as this is not actually any new functionality. Now these tests just repeat!

This might however be another point where we all think very differently. I honestly can't even do a review/look at the tests now, because everything is duplicated and I fiercely, fiercely hate duplication. \*angry cat noises\*

frontend/cypress/integration/actions_spec.ts
```diff
@@ -387,18 +494,35 @@ const createDeleteSeed = () => {
...
     const seed = new EvaluationSeed({
         progression: faker.random.arrayElement(Object.values(Progression)),
         users: getUsers(faker.datatype.number({ min: 1, max: 5 })),
     })
 
+    const editor = faker.random.arrayElement(seed.participants)
+
     const actions: Action[] = Array.from({ length: faker.datatype.number({ min: 2, max: 4 }) }, () => {
-        return seed.createAction({ questionOrder: faker.datatype.number({ min: 1, max: 8 }) })
+        return seed.createAction({ questionOrder: faker.datatype.number({ min: 1, max: 8 }), assignedTo: editor })
```
achaikou, line 521: With setting "assignedTo" you kill randomization in tests that do not require it

frontend/cypress/integration/actions_spec.ts
```diff
@@ -240,85 +294,130 @@ describe('Actions', () => {
...
 
-        it('Completing Action can be cancelled', () => {
-            actionsGrid.actionLink(existingAction.questionOrder, existingAction.title).click()
-            checkView()
-            openConfirmCompleteView()
-            editActionDialog.completeActionCancelButton().click()
-            checkView()
-            editActionDialog.assertNoClosingMessageInNotes(existingNotes)
+        it('Action can be completed', () => {
+            const saveWithComment = Math.random() < 0.5
```
achaikou, line 404: or `faker.datatype.boolean()` which uses some better uniform randomization

frontend/cypress/integration/actions_spec.ts
```diff
@@ -240,85 +294,130 @@ describe('Actions', () => {
...
+            const { updatedAction: completedAction, newNote } = createCompleteActionData(seed, editor, existingAction)
+
+            editActionDialog.assertViewActionNotCompleted()
```
achaikou, line 407: Even if you would not agree with me that we need to bring down the file size and basically unify all the complete and edit tests into one, can you at least do not check the views every single time?
Tests must be as slim as possible, meaning that you do not check stuff twice.

As I wrote in the comments to the previous review, viewing and checking overlay is one thing. Just completing action requires waaaaaaaay less steps.

frontend/cypress/integration/actions_spec.ts
achaikou: Disagree. Unify all the tests and remove the view check.

frontend/cypress/integration/actions_spec.ts
achaikou: By "reasonably" I meant with regards to possible dates which in my ideal world won't be today, date formats and time zones so we do not end up in a bad place :smile: 

but if something will work in a stable way that is easy to bulk-update, ok :smile:

frontend/cypress/integration/actions_spec.ts
achaikou: Agree with Petter that waiting time is for some reason bigger for notes/completing than for everything else, and we need to investigate why, but if it is still time, then instead of doing static wait you simply do dynamic wait inside your "complete" function - as it is part of it.
`cy.get("my_complete_view_or_whatever_there_is", { timeout: 8000 }).should("not.exist")`
This way we will simply increase timeout from default cypresses 4 seconds to whatever we need (8?), and it won't be static. So we won't have to wait the full amount of time if action completes faster.

frontend/cypress/integration/actions_spec.ts
achaikou: Need to create Contributing Guidelines, where we all agree on something.
I decided that "get" makes no sense for locators and selectors. You will always "get" them.

Basically there was supposed to be a difference between getting the selector and methods (like assert and so on), but I imagined majority of things will be selectors, and very few methods like "assert". Looks like that didn't happen. Maybe we need a better way to divide selectors from methods.

achaikou (commented):
Can't do a proper review as it's too crowded here already, and all of us have very different opinions about the testing framework (we might need to add contributing guidelines), so just a couple of comments that I believe I didn't try fighting anyone else before.

frontend/cypress/integration/actions_spec.ts
denektenina: Thanks for the advise everybody! I have now submitted a new solutions without wait. And at least it runs :D

frontend/cypress/integration/actions_spec.ts
denektenina: Do you think I should do something here, though, @achaikou ?

achaikou: I think you can leave it as is. It's not the main feature we are expected to test.

Like you will definitely get into small trouble with time - it's not unreasonable to think that notes will be created in the previous minute or next minute, not in the current one. So you at least would have to check that time equals A or A-1 or A+1 and I am not sure if or how this could be done with cypress. Like there are some  "within" assertions, but I am not sure how well they will work, especially on string dates...

Date should be a bit easier, but if we schedule the runs around midnight (or someone who cannot sleep will do them then), you fall under the same trouble, but now with dates. (And if someone runs precisely at the daylight change - :smile: . Yep, unlikely :smile: )

So go on if you want to do it, but I would say leave it be for now if you don't. 
(I was burned by testing dates, timezones, daylight saving and date formats before, didn't like it, but now situation is different, so I have a hope this will definitely go way better.)

denektenina: Ok, leaving it since it's not really part of my PR :)

frontend/cypress/integration/actions_spec.ts
denektenina: Trying Alenas solution since I agree that we get the same checks using randomness (sometimes with and sometimes without comment).

frontend/cypress/integration/actions_spec.ts
hnformentin: Great. It had problems in the pipeline, but I could not identify if it is related to it:
https://dashboard.cypress.io/projects/ry3x7y/runs/106/test-results/0a8136bc-5a50-4d0b-b50e-fbf606065a7b

frontend/cypress/integration/actions_spec.ts
hnformentin: The suggestion was indeed to only check date and hours, but not minutes. But fine the way you decided.

frontend/cypress/integration/actions_spec.ts
hnformentin: Good point. I thought that having `get` would make easier to identify selectors and locators, but indeed, it can become repetition. Good idea to have a Contributing Guidelines 🥇

frontend/cypress/integration/actions_spec.ts
denektenina: @hnformentin Can you please click "resolve conversation" when you are happy with an answer or a change? That way I don't have to read the same comments every time I go through the PR :)

frontend/cypress/integration/actions_spec.ts
denektenina: Can you check if this is still relevant after my refactoring, @hnformentin?

frontend/cypress/integration/actions_spec.ts
denektenina: I think this is fixed now that I removed WAIT, right?

frontend/cypress/integration/actions_spec.ts
denektenina: Fixed, please resove @petterwildhagen

frontend/cypress/integration/actions_spec.ts
denektenina: Can you see if this is still relevant, @hnformentin ?

frontend/cypress/integration/actions_spec.ts
denektenina: This feels too big for my PR, @hnformentin. Please resolve if you agree :D

frontend/cypress/integration/actions_spec.ts
denektenina: I think this is done now. Please resolve if you agree @petterwildhagen, or elaborate if not :)

frontend/cypress/support/action.ts
```diff
@@ -151,6 +205,8 @@ export class EditActionDialog extends ActionDialog {
         this.assignedToInput().should('have.value', action.assignedTo.user.name)
         this.priorityInput().should('have.text', mapPriority(action.priority))
         this.dueDateInput().should('have.value', action.dueDate.toLocaleDateString(FUSION_DATE_LOCALE))
+        this.actionCompletedText().should(action.completed ? 'exist' : 'not.exist')
+        this.completeActionButton().should(action.completed ? 'not.exist' : 'exist')
```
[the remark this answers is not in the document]
denektenina: Fixed

frontend/cypress/integration/actions_spec.ts
denektenina: After our talk yesterday I have gone through and reassessed  my checks, so I think I can defend the checks I am doing now, and have removed others. Please review again and resolve if you are happy :)

frontend/cypress/integration/actions_spec.ts
denektenina: Yes, I need it in the tests that are testing the action table that only displays the actions assigned to the logged in user. I think I removed it from the other places now.

frontend/cypress/integration/actions_spec.ts
denektenina: Fixed

frontend/cypress/integration/actions_spec.ts
denektenina: Fixed, I think - please check if you agree with my choice now

frontend/cypress/integration/actions_spec.ts
hnformentin: Fine for me if it is fine for both you!

frontend/cypress/integration/actions_spec.ts
hnformentin: I thought as it was your decision you could resolve it, but I can do as well ;-)

frontend/cypress/support/action.ts
```diff
@@ -162,17 +230,67 @@ export class EditActionDialog extends ActionDialog {
...
         })
     }
 
+    /**
+     * Check that the provided closing note is in the provided index.
+     */
+    assertClosingMessageInNotes = (note: Note, index: number) => {
+        this.notesDiv().children().eq(index).as('note')
+        cy.get('@note').should('contain.text', note.createdBy.user.name + ' closed action')
+        if (note.text) {
```
achaikou, line 243: Once again: please remember that you never actually check this text https://github.com/equinor/fusion-bmt/pull/543#discussion_r700360967

achaikou (commented):
You already fixed the seed.plant instability, so good :+1: 
The failing tests are probably part of the instability I am currently investigating. Can't say anything yet.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -83,8 +84,7 @@ describe('Actions', () => {
         const actionProgression = faker.random.arrayElement([Progression.Workshop, Progression.FollowUp])
 
         beforeEach(() => {
-            ;({ seed, existingAction, existingNotes } = createEvaluationWithActionsWithNotes({ title: 'Just some random action' }))
-            editor = faker.random.arrayElement(seed.participants)
+            ;({ seed, editor, existingAction, existingNotes } = createEvaluationWithActionsWithNotes({ title: 'Just some random action' }))
```
petterwildhagen, line 87: replace editor with user and login in the same way as other tests in actions_spec.
only return seed in before each.
Move getting of existingAction & existingNotes into test itselg

frontend/cypress/integration/actions_spec.ts
```diff
@@ -96,18 +96,9 @@ describe('Actions', () => {
...
-            dropdownSelect.select(editActionDialog.assignedToInput(), updatedAction.assignedTo.user.name)
-            editActionDialog.dueDateInput().replace(updatedAction.dueDate.toLocaleDateString(FUSION_DATE_LOCALE))
-            dropdownSelect.select(editActionDialog.priorityInput(), mapPriority(updatedAction.priority))
-            editActionDialog.descriptionInput().replace(updatedAction.description)
-            newNotes.forEach(note => {
-                editActionDialog.notesInput().replace(note.text)
-                editActionDialog.addNoteButton().click()
-            })
-
+            editActionDialog.editAction(updatedAction, newNotes, dropdownSelect)
```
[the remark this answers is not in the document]
denektenina: Fixed

frontend/cypress/integration/actions_spec.ts
denektenina: If this is still relevant in the newest PR, please show me where the duplicates are. You say I "same test 3 times", is this in the complete-tests or other places. If in the complete tests, we can fix this for my new PR. If not, we can fix in a later PR.

If you are referring to lines like: 

```
editActionDialog.checkIfCompletedInView(action.completed)
editActionDialog.completeActionButton().click()
editActionDialog.checkIfCompleteConfirmViewVisible(true)
```

being repeated in the same order several places, please specify if you argue for **removing** some of the lines because it is not needed to check, or if you argue for **placing them in a separate function and reusing** this to avoid code duplicate. 

What I meant in my comment above is that I did unify the "complete with" and "complete without" comment-tests. This removed a lot of duplicates. I think cancel complete differs, but please argue if you think something should be changed there. If this is too much to explain to me in writing, we could maybe have a call and make the changes together?

frontend/cypress/support/action.ts
```diff
@@ -162,17 +230,67 @@ export class EditActionDialog extends ActionDialog {
...
+            cy.get('@note').contains(note.text).should('exist')
+        }
+    }
+
+    /**
+     * Check that the notes list does not contain a closing note
```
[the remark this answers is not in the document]
denektenina: I think this is fixed in the new PR

frontend/cypress/support/action.ts
denektenina: I think this is fixed in the new PR

frontend/cypress/integration/actions_spec.ts
denektenina: Basing my new PR on your newest PR so I think this is fixed

frontend/cypress/integration/actions_spec.ts
achaikou: Thanks for trying to explain this! :heart: 

New PR at the moment doesn't introduce more tests, so it is not relevant to it. And currently I am happy with the "check" functions. And in general I like new PR very much :heart: 

By duplicates I mean 
- complete/edit action from workshop/follow up
- complete/edit action from Actions Tab on Follow Up
- complete/edit action from Dashboard

The flow of these actions is repeated. The meaning of these tests is repeated. That's what I don't like.

There were 3 options for me how to deal with that
1. Randomization (you randomize the way you enter the dialog) - you run it just once from random entry point
2. For loop, like for roles - you run it 3 times, for each way to enter the dialog (as with randomization, the whole flow is still covered by 1 code sequence and 3 entry functions). I do not like this approach, as it will take long to complete and doesn't introduce much value to the tests. 
3. wait until we establish this "lightweight second level" of Petters test strategy where we will only check from all 3 places that 1) edit link is present 2) Complete action button is present 3) whatever else we need to check (that notes number is as expected?) without actually doing or checking anything else

I am at the moment for a mix of approaches 1 and 3, but that requires Petters vision implemented as example for some tests and our discussion on randomization to happen

PR #594, opened 2021-09-03
Add Role-based restriction to backend mutations

### LINKED ISSUE #512: Safeguard the backend against the last 2 facilitators being able to delete each other
Looks like we have the following problem: not only facilitators can access the evaluation in Nomination stage, but everyone.

- If the person isn't added to evaluation, they need the direct link to access it. But then they are free to do anything with users.
- If the person is added to evaluation as participant, they already get the link to evaluation on their dashboard, thus can go directly into it and do any wicked things they want to.

Noticed during quick testing, so more related problems can actually exist.

### LINKED ISSUE #537: Restric mutation of data in graphql
Related to #469

ErlendHaa:
This PR adds role-based restrictions to all exposed mutations. Additionally it extends the current test-framework to test the new restrictions.

The test framework for Mutation testing in the backend only implements access-based tests at the moment, but it is intended to be used for further correctness testing of the mutations. However that is out of scope of this PR.

Closes #537
Closes #512

### DIFF
--- a/backend/api/Authorization/AuthService.cs
+++ b/backend/api/Authorization/AuthService.cs
@@ -9,7 +9,6 @@ namespace api.Authorization
     public interface IAuthService
     {
         public string GetOID();
-        public void AssertIsFacilitator(string evaluationId);
     }
 
     public class AuthService : IAuthService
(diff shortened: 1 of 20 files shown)


backend/api/Authorization/AuthService.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
```
ReedOnly, line 2: probably don't need linq here

backend/api/GQL/Mutation.cs
```diff
@@ -209,13 +262,10 @@ public Note CreateNote(string actionId, string text)
...
 
-            return _noteService.Create(CurrentUser(evaluation), text, action);
-        }
+            Role[] canBePerformedBy = { Role.Facilitator, Role.Participant, Role.OrganizationLead };
+            AssertCanPerformMutation(evaluation, canBePerformedBy);
 
-        public Note EditNote(string noteId, string text)
-        {
-            Note note = _noteService.GetNote(noteId);
-            return _noteService.EditNote(note, text);
```
ReedOnly, line 218: I guess we are not using, nor want to have the possibility to edit Notes?

ReedOnly (approved):

backend/api/GQL/Mutation.cs
ErlendHaa: That's my so-far conclusion, yes.

>This mutation is unused and there are currently no UI support for
editing notes - nor is any such feature planned.

Updating and maintaining this mutation without any current or future planned use-case for it seams like unnecessary work. If we ever need it, we can revert this commit

PR #622, opened 2021-09-09
Remove logging from plant method

petterwildhagen:


ErlendHaa:
Cleaning up the log is a good idea. But this PR removes _all_ the relevant parts of the log, but leaves _all_ unrelated third-party logging. We used the now removed logs just yesterday to figure out the buggy test, so they definitely have debugging value. If anything they should be _enriched_. 

I suggest you aim to remove all these logs instead:

- (req fn)  GET https://pro-s-context-ci.azurewebsites.net/.*
- (req stub) GET https://pro-s-portal-ci.azurewebsites.net/.*
- (req fn) GET https://pro-s-people-ci.azurewebsites.net/.*
- POST 200 http://localhost:5000/graphql

These give little to none debugging value and combined they are responsible for approx half of all the log entries. Removing them would significantly reduce the noise in the log. The cypress docs explains how to suppress third party log entries

petterwildhagen:
Closing this PR and setting task back to to-do.

PR #631, opened 2021-09-13
Remove default arg for role and progression from EvaluationSeed, place randomization of test data in test

petterwildhagen:


frontend/cypress/support/evaluation_seed.ts
```diff
@@ -84,10 +83,8 @@ export class EvaluationSeed {
         this.progression = progression
         let participants: Participant[] = []
 
-        if (roles === undefined) {
-            users.forEach(u => {
-                participants.push(this.createParticipant({ user: u, progression: progression }))
```
achaikou, line 89: Now that you made both roles and progression obligatory, does `createParticipant` function give any value?
It's used once and its presence only creates confusion on your intended way of creating participants setup.

frontend/cypress/support/evaluation_seed.ts
```diff
@@ -84,10 +83,8 @@ export class EvaluationSeed {
...
-            })
+        if (roles.length !== users.length) {
+            throw new Error('EvaluationSeed: roles.length != users.length')
         } else {
```
achaikou, line 88: You don't need else anymore as throw works as a guard

frontend/cypress/integration/actions_spec.ts
```diff
@@ -425,19 +441,29 @@ const createEditTestData = (seed: EvaluationSeed, user: User, existingAction: Ac
...
+            }),
         })
     })
 
     return { updatedAction, newNotes }
 }
 
 const createCompleteAction = (user: User, existingAction: Action) => {
+    const assignableRoles = [Role.Participant, Role.Facilitator, Role.OrganizationLead]
+    const progressions = [Progression.Individual, Progression.FollowUp, Progression.Preparation, Progression.Workshop]
```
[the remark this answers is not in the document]
petterwildhagen: No particular reason. I just did not include it - because as you say it looks weird. But I can fix it.
Fixed

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: Good point. Fixed

frontend/cypress/support/evaluation_seed.ts
petterwildhagen: No. Deleted.

frontend/cypress/integration/actions_spec.ts
```diff
@@ -454,7 +454,13 @@ const createEditTestData = (seed: EvaluationSeed, user: User, existingAction: Ac
 
 const createCompleteAction = (user: User, existingAction: Action) => {
     const assignableRoles = [Role.Participant, Role.Facilitator, Role.OrganizationLead]
-    const progressions = [Progression.Individual, Progression.FollowUp, Progression.Preparation, Progression.Workshop]
+    const progressions = [
+        Progression.Individual,
```
achaikou, line 458: a) You put Individual twice
b) I meant not only in this line, but in all 3 places you changed in the code.
c) You as well can go with `Object.values(Progression)`

petterwildhagen: Fixed (in all places). Using Object.values(Progression)

achaikou (approved):
:cat: :+1:

PR #642, opened 2021-09-15
Made view for editable question

denektenina:
It is now possible to edit a question template in the admin view.

image

**This PR features:**
* Improved rendering of questions with markdown
* Validation of text field
* Error message when saving question template fail

**Some implementation choices:**
* I reused some code from `ActionEditForm.tsx,` so I extracted it into a helper function and used it both in the new and old place
* I removed `BarrierQuestionList.tsx `because it didn't do much, and just cased an extra level to send parameters through
* I moved the view code from AdminQuestionItem and moved it to a new file StaticQuestionItem, so that which view is rendered is depending on whether or not we are in edit mode.


frontend/src/api/fragments.ts
```diff
@@ -65,6 +65,7 @@ export const QUESTIONTEMPLATE_FIELDS_FRAGMENT = gql`
         barrier
         order
         organization
+        status
```
denektenina, line 68: This is needed for when I change a questionTemplate (this value must be sent in)

frontend/src/api/fragments.ts
```diff
@@ -95,15 +96,15 @@ export const ACTION_FIELDS_FRAGMENT = gql`
...
-    actions {
-        ...ActionFields
-        notes {
-            id
+    fragment QuestionActions on Question {
+        actions {
+            ...ActionFields
+            notes {
+                id
+            }
```
denektenina, line 104: This is just indentation from prettier.

frontend/src/components/Action/EditForm/ActionEditForm.tsx
```diff
@@ -111,27 +112,11 @@ const ActionEditForm = ({
...
     useEffect(() => {
-        if (assignedToValidity === 'error') {
-            if (checkIfParticipantValid(assignedTo)) {
-                setAssignedToValidity('success')
-            }
-        } else if (assignedToValidity === 'success') {
-            if (!checkIfParticipantValid(assignedTo)) {
-                setAssignedToValidity('error')
-            }
-        }
```
denektenina, line 134: Extracted code into helper for reuse

frontend/src/views/Admin/AdminQuestionItem.tsx
```diff
@@ -1,41 +1,42 @@
...
-                    </Box>
-                </Box>
-                <Box mr={2}>
-                    <EditIcon cursor="pointer" color={tokens.colors.interactive.primary__resting.rgba} />
-                </Box>
-                <Box>
-                    <MoreIcon cursor="pointer" color={tokens.colors.interactive.primary__resting.rgba} />
-                </Box>
-            </Box>
+            {isInEditmode ? (
```
denektenina, line 29: Moved this to new file since we want to render a different view if we are in edit mode

frontend/src/views/Admin/BarrierQuestionList.tsx
```diff
@@ -1,20 +0,0 @@
...
-    return (
-        <div>
-            {orderedQuestions.map(q => {
-                return <AdminQuestionItem key={q.id} question={q} />
-            })}
-        </div>
-    )
-}
-
-export default BarrierQuestionList
```
denektenina, line 20: Moved this logic to the component above in the tree, since it doesn't add much logic or view logic

frontend/src/views/Admin/EditableQuestionItem.tsx
```diff
@@ -0,0 +1,141 @@
+import React, { useEffect, useState } from 'react'
```
denektenina, line 1: This file has most of the logic in this PR:

* Renders new edit question template-view
* Validates form
* Renders error message
* Triggers save question when user clicks save
* Displays spinner while saving

frontend/src/views/Admin/QuestionListWithApi.tsx
```diff
@@ -1,17 +1,30 @@
...
+        loading: isQuestionTemplateSaving,
+        questionTemplate,
+        error: questionTemplateSaveError,
+    } = useQuestionTemplateMutation()
+
+    useEffectNotOnMount(() => {
+        if (!isQuestionTemplateSaving) {
+            refetchQuestionTemplates()
+        }
+    }, [isQuestionTemplateSaving])
```
denektenina, line 27: Backend doesn't return the same questionTemplate after a questionTemplate is saved, but instead returns a new question template (with a new ID), and sets the old question template to INACTIVE. This caused problems when trying to cache the result, because it was not possible to replace the existing instance in the cache, since this had a different ID. It was also not possible to just correct the old instance (setting state to INACTIVE) and adding the new question template, because for some reason, Apollo was only able to return the same amount of cache instances that was there previously, leaving out the newly added instance.

Therefore, instead of manually changing the cache, we are now **refetching** the question template data after a question template has been edited. This is a good solution because it only causes one additional call every time we save a question template, and it ensures the data is synchronized, minimizing the chance of error caused by manually updating the cache.

maiaht (approved):

PR #649, opened 2021-09-19
Made menu for question template

denektenina:
It now looks like this when you click the menu-button:

image

**Some notes:**
* I had to upgrade the equinor-design-system-package from 0.8 to 0.13 to get a Menu-component that worked. We should check if it's ok to merge that. Because of the package change I had to make a change to the naming of some tab-elements, Tooltip, and some small styling changes (margins that suddenly were lacking). That is actually most of the files in this PR.
* The buttons in the menu are not yet implemented, as that was not part of the issue.
* The menu gets a strange dashed line around the menu element when an element is clicked. I think this has to do with keyboard-control / accessability, but I don't think it's possible to remove (and anyways we probably shouldn't)?

image

### DIFF
--- a/frontend/src/views/Admin/StaticQuestionItem.tsx
+++ b/frontend/src/views/Admin/StaticQuestionItem.tsx
@@ -1,11 +1,12 @@
-import React from 'react'
-import { Chip, EditIcon, MarkdownViewer, MoreIcon } from '@equinor/fusion-components'
-import { Typography } from '@equinor/eds-core-react'
-import { tokens } from '@equinor/eds-tokens'
+import React, { useRef, useState } from 'react'
+import { Chip, MarkdownViewer } from '@equinor/fusion-components'
+import { Button, Icon, Typography } from '@equinor/eds-core-react'
 import { Box } from '@material-ui/core'
+import { more_vertical, edit } from '@equinor/eds-icons'
 
 import { QuestionTemplate } from '../../api/models'
 import { organizationToString } from '../../utils/EnumToString'
+import QuestionTemplateMenu from './QuestionTemplateMenu'
 
 interface Props {
     question: QuestionTemplate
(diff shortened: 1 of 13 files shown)


frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -13,27 +13,69 @@ interface Props {
...
+                    </Button>
+                    <Menu
+                        id="menu-complex"
+                        open={isOpen}
+                        anchorEl={anchorRef.current as HTMLButtonElement}
+                        onClose={closeMenu}
+                        placement={'bottom'}
+                    >
+                        <Menu.Item onClick={() => {}}>
+                            <Icon data={arrow_up} size={16} />
```
denektenina, line 58: I will make a new component for the menu

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -1,8 +1,8 @@
-import React from 'react'
-import { Chip, EditIcon, MarkdownViewer, MoreIcon } from '@equinor/fusion-components'
-import { Typography } from '@equinor/eds-core-react'
-import { tokens } from '@equinor/eds-tokens'
+import React, { useRef, useState } from 'react'
+import { Chip, MarkdownViewer } from '@equinor/fusion-components'
+import { Button, Icon, Menu, Typography } from '@equinor/eds-core-react'
 import { Box } from '@material-ui/core'
+import { arrow_up, arrow_down, delete_to_trash, more_horizontal, edit } from '@equinor/eds-icons'
```
maiaht, line 5: Why not use the icon `more_vertical` which looks like the icon in the design?

denektenina: Yes, I can do that

denektenina: Fixed

frontend/src/views/Admin/StaticQuestionItem.tsx
denektenina: Fixed

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -1,39 +1,57 @@
...
-                    </div>
-                    <MarkdownViewer markdown={question.supportNotes} />
+                    <Button variant="ghost" color="secondary" onClick={() => setIsInEditmode(true)}>
+                        <Icon data={edit}></Icon>
+                    </Button>
+                </Box>
+                <Box>
+                    <Button variant="ghost" color="secondary" ref={anchorRef} onClick={() => (isOpen ? closeMenu() : openMenu())}>
+                        <Icon data={more_vertical}></Icon>
+                    </Button>
```
maiaht, line 50: Should the colors of these buttons be "primary" instead of "secondary"? And should they be aligned with the icons (the plus icon and the more icon) in the bar above?

denektenina: Yes, I agree that that's better :) Fixed

denektenina: Did not align, because it has to be buttons (well, the menu icon has to be button to follow the design pattern of the Menu-component), so they will take up extra space because of that. Later, we might change the other icons to buttons as well (especially the menu icon on the top level), and then they will probably be aligned again

maiaht (approved):

PR #650, opened 2021-09-20
Add dropdown to select project category in Admin view

maiaht:


denektenina (approved):
Looks good! Ran it locally and it looks good, but I couldn't really test it since all the questions belong to the same category. Also haven't run tests, but I guess that's done automatically.

PR #661, opened 2021-09-21
Add new question template

maiaht:
The goal of this PR is to make it possible to add a new question template to a barrier. The component that is used to do this is quite similar to the component `EditableQuestionItem`, and consequently some refactoring has been done to avoid code duplication. Since a new question template does not belong to any project categories, the mutation that is used to create a question template cannot use the existing fragment for question templates, and consequently a new fragment has been made.

### DIFF
--- a/frontend/src/api/fragments.ts
+++ b/frontend/src/api/fragments.ts
@@ -70,16 +70,13 @@ export const QUESTIONTEMPLATE_FIELDS_FRAGMENT = gql`
     }
 `
 
-export const QUESTIONTEMPLATE_PROJECT_CATEGORY_FIELDS_FRAGMENT = gql`
-    fragment QuestionTemplateProjectCategoryFields on QuestionTemplate {
-        ...QuestionTemplateFields
+export const PROJECT_CATEGORY_FIELDS_FRAGMENT = gql`
+    fragment ProjectCategoryFields on QuestionTemplate {
         projectCategories {
             id
             name
         }
-        __typename
     }
-    ${QUESTIONTEMPLATE_FIELDS_FRAGMENT}
 `
 
 export const QUESTION_ANSWERS_FRAGMENT = gql`
--- a/frontend/src/api/fragments.ts
+++ b/frontend/src/api/fragments.ts
@@ -66,12 +66,20 @@ export const QUESTIONTEMPLATE_FIELDS_FRAGMENT = gql`
         order
         organization
         status
+        __typename
+    }
+`
+
+export const QUESTIONTEMPLATE_PROJECT_CATEGORY_FIELDS_FRAGMENT = gql`
+    fragment QuestionTemplateProjectCategoryFields on QuestionTemplate {
+        ...QuestionTemplateFields
         projectCategories {
             id
             name
         }
         __typename
     }
+    ${QUESTIONTEMPLATE_FIELDS_FRAGMENT}
 `
 
 export const QUESTION_ANSWERS_FRAGMENT = gql`
--- a/frontend/src/views/Admin/AdminView.tsx
+++ b/frontend/src/views/Admin/AdminView.tsx
@@ -1,7 +1,6 @@
-import React from 'react'
+import { useRef, useState } from 'react'
 import { Button, Divider, Icon, Typography } from '@equinor/eds-core-react'
 import { add, more_vertical } from '@equinor/eds-icons'
-import { useRef, useState } from 'react'
 import { ApolloError, gql, useQuery } from '@apollo/client'
 import { SearchableDropdown, SearchableDropdownOption, TextArea } from '@equinor/fusion-components'
 import { Box } from '@material-ui/core'
@@ -12,6 +11,7 @@ import QuestionListWithApi from './QuestionListWithApi'
 import { barrierToString } from '../../utils/EnumToString'
 import { apiErrorMessage } from '../../api/error'
 import BarrierMenu from './BarrierMenu'
+import CreateQuestionItem from './CreateQuestionItem'
 
 interface Props {}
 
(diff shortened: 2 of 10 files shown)


frontend/src/views/Admin/CreateQuestionItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    useEffectNotOnMount(() => {
+        if (!isTextfieldValid()) {
+            setTextValidity('error')
+        }
+    }, [text])
+
+    useEffect(() => {
+        updateValidity(isTextfieldValid(), textValidity, setTextValidity)
+    }, [text, textValidity])
```
denektenina, line 41: Seems like this validation code isn't used anywhere?

frontend/src/api/fragments.ts
```diff
@@ -66,12 +66,20 @@ export const QUESTIONTEMPLATE_FIELDS_FRAGMENT = gql`
         order
         organization
         status
+        __typename
+    }
+`
+
+export const QUESTIONTEMPLATE_PROJECT_CATEGORY_FIELDS_FRAGMENT = gql`
+    fragment QuestionTemplateProjectCategoryFields on QuestionTemplate {
+        ...QuestionTemplateFields
```
denektenina, line 75: This is a bit chuncky (the name is very long and maybe not very accurate?). I would suggest to instead have one fragment for QUESTIONTEMPLATE_FIELDS_FRAGMENT (like you do) and then one for only 
projectCategories {
            id
            name
        }


and then importing them both where you need them booth instead of enforcing all questionTemplate fields whenever you need projectCategoryFields

frontend/src/views/Admin/QuestionListWithApi.tsx
```diff
@@ -89,10 +89,10 @@ const useQuestionTemplatesQuery = (): QuestionTemplatesQueryProps => {
     const GET_QUESTIONTEMPLATES = gql`
         query {
             questionTemplates (where: {status: {eq: ${Status.Active}} }) {
-                ...QuestionTemplateFields
+                ...QuestionTemplateProjectCategoryFields
             }
         }
-        ${QUESTIONTEMPLATE_FIELDS_FRAGMENT}
+        ${QUESTIONTEMPLATE_PROJECT_CATEGORY_FIELDS_FRAGMENT}
```
denektenina, line 95: This name is difficult to understand

denektenina (commented):
Looks good, except for my few comments :)

frontend/src/api/fragments.ts
maiaht: Good idea! Fixed

frontend/src/views/Admin/CreateQuestionItem.tsx
maiaht: It seems this code actually is used, but not exactly how we wanted. The validity of the text is only checked after you have started to write something, but has now been changed, and it is not possible to save a question if the title is empty.

frontend/src/views/Admin/QuestionListWithApi.tsx
maiaht: Fixed

denektenina (commented):
Something doesn't feel right for me regarding where the comment winds up after being created. After "Save" is completed, there isn't really any confirmation that the save was successful, maybe it could be possible to scroll it into view (automatically)? 

Also, I don't know what we can do about this, but I think it's a bit strange that the comment is added to for instance the "General"-barrier, but with a number that is far from the other numbers in the barrier (for instance the first questions are 1-37 and then the new one, that is placed right beneath is number 107).

frontend/src/views/Admin/Components/CancelOrSaveQuestion.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const CancelOrSaveQuestion = ({ isQuestionTemplateSaving, setIsInMode, onClickSave, questionTitle }: Props) => {
+    return (
+        <>
+            {isQuestionTemplateSaving && (
+                <Box alignSelf={'flex-end'} mb={1}>
+                    <SaveIndicator savingState={SavingState.Saving} />
+                </Box>
+            )}
+            <Box alignSelf={'flex-end'}>
```
denektenina, line 21: I see that there is some strange placement of the cancel and save-buttons when there are lots of text in the edit-view. 

image

I fixed it in my PR, but maybe you lost it when merging. I think I just did `style={{ minWidth: '170px'}}` on this line.

frontend/src/views/Admin/CreateQuestionItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+                            menuItems={['strong', 'em', 'bullet_list', 'ordered_list', 'blockquote', 'h1', 'h2', 'h3', 'paragraph']}
+                        >
+                            {supportNotes}
+                        </MarkdownEditor>
+                    </Box>
+                </Box>
+                <Box display="flex" flexDirection={'column'}>
+                    <Box flexGrow={1}>
+                        <SearchableDropdown
+                            label="Responsible discipline"
```
denektenina, line 92: This should be called "organization" instead of "Responsible discipline"

frontend/src/views/Admin/AdminView.tsx
```diff
@@ -20,8 +20,10 @@ const AdminView = ({}: Props) => {
     const [selectedProjectCategory, setSelectedProjectCategory] = useState<string>('all')
     const [isMenuOpen, setIsMenuOpen] = useState<boolean>(false)
     const [isInAddCategoryMode, setIsInAddCategoryMode] = useState<boolean>(false)
+    const [isAddingQuestion, setIsAddingQuestion] = useState<boolean>(false)
     const headerRef = useRef<HTMLElement>(null)
     const menuAnchorRef = useRef<HTMLButtonElement>(null)
+    const quesstionTitleRef = useRef<HTMLElement>(null)
```
denektenina, line 26: Typo, should not be two s'es :)

frontend/src/views/Admin/Components/CancelOrSaveQuestion.tsx
maiaht: Fixed

frontend/src/views/Admin/CreateQuestionItem.tsx
maiaht: Fixed

frontend/src/views/Admin/AdminView.tsx
maiaht: Fixed

denektenina (approved):
There is a misalignment between the text and the organization-box, but other than that, looks good!

PR #662, opened 2021-09-21
Add cypress test for edit question template

petterwildhagen:


frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,17 +1,44 @@
...
 describe('Admin page', () => {
-    before(() => {
+    beforeEach(() => {
         const user = users[users.length - 1]
         cy.visitProject(user)
     })
 
-    it('Navigate to Admin page', () => {
+    it.only('Edit question - change title, support notes and  ', () => {
+        const questionNo = faker.random.arrayElement([1, 2, 3, 4])
```
petterwildhagen, line 11: This is a 'static' assumption that we have 4 questions on the page. Should be improved to either count the questions on the page or somehow dynamically link to the initially loaded test data created in InitContent.cs

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,17 +1,51 @@
...
+    beforeEach(() => {
         const user = users[users.length - 1]
         cy.visitProject(user)
     })
 
-    it('Navigate to Admin page', () => {
-        cy.get('button').contains('Admin').click()
-        cy.contains('Project configuration: Questionnaire').should('be.visible')
-    })
+    it.only('Edit question - change title, support notes and  ', () => {
```
maiaht, line 11: Is it supposed to be `it.only` here? Also, the title of this test does not seem to be finished

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -33,10 +33,15 @@ const StaticQuestionItem = ({ question, setIsInEditmode }: Props) => {
...
+                            {question.text}
+                        </Typography>
                         <Box display="flex" flexDirection="row" mb={3} mt={1}>
                             <Box mr={1}>
-                                <Chip style={{ backgroundColor: tokens.colors.infographic.primary__spruce_wood.rgba }}>
+                                <Chip
+                                    style={{ backgroundColor: tokens.colors.infographic.primary__spruce_wood.rgba }}
+                                    data-testid={'organization-' + question.order}
+                                >
                                     <Tooltip title={'Responsible discipline'} placement={'bottom'}>
```
maiaht, line 45: Should "Responsible discipline" be changed to "Organization" here as well?

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,17 +1,51 @@
...
+            cy.getByDataTestid('save-question').click()
+            cy.getByDataTestid('question-title-' + questionNo).should('have.text', newTitle)
+            cy.getByDataTestid('organization-' + questionNo).should('have.text', organization)
+            cy.getByDataTestid('question-title-' + questionNo)
+                .parent()
+                .within(() => {
+                    cy.contains(newSupportNotes).should('be.visible')
+                })
+        })
+    })
```
maiaht, line 46: Should the "Cancel"-button also be tested?

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: No. Fixed. Completed test title (the test is finished I fogot to add 'organization' to the title)

frontend/src/views/Admin/StaticQuestionItem.tsx
petterwildhagen: Good catch. Fixed

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Yes it should. Added

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+        .shadow()
+        .within(() => {
+            cy.get('[id=editor]').replace(newSupportNotes)
+        })
+    cy.get('label')
+        .contains('Organization')
+        .next()
+        .click()
+        .clear()
+        .type(organization + '{enter}')
```
achaikou, line 77: use `replace` again?

Other thing: consider if you want this behave closer to the user experience?

This is a select, so users will unlikely type stuff. Instead they will choose from the list. So maybe you would like to do
```
    cy.get(`#question-${questionNo}`).within(() => {
        // put organizationSelect = cy.contains('Organization').next() into admin.ts support file
        new DropdownSelect().select(adminQuestions.organizationSelect(), organization)
    })
```

(if we want to use "within" we need to update the select methods such as all `cy.get("section")` are surrounded by any escape workaround because "section" appeares in totally different part of the dom

```
        selectInput.click()
        cy.document()
            .its('body')
            .within(() => {
                cy.get('section').contains('span', text).click()
            })
```

If you prefer to bypass within, then it shouldn't be needed.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+    cy.get('label').contains('Organization').next().invoke('attr', 'value').should('eq', organization)
```
achaikou, line 78: Why do you need this one? You check organization after save, is that not enough?

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+            cy.getByDataTestid('organization-' + questionNo).then(function (t) {
+                let currentOrg = Cypress.$(t).text()
+                let organization = faker.random.arrayElement(Object.keys(Organization).filter(t => t !== currentOrg))
+
+                changeQuestionFields(questionNo, newTitle, newSupportNotes, organization)
+                cy.getByDataTestid('save-question').click()
+                cy.getByDataTestid('question-title-' + questionNo).should('have.text', newTitle)
+                cy.getByDataTestid('organization-' + questionNo).should('have.text', organization)
+                cy.getByDataTestid('question-title-' + questionNo)
+                    .parent()
```
achaikou, line 29: Maybe limit scope instead by
```cy.get(`#question-${questionNo}`).within(() => {```

Explicitly stating that "I am interested in text that is located inside question" seems more readable than relying on the parent. Especially if all these locators are in the test code, and not in support files ;(

File: frontend/cypress/integration/admin_page_spec.ts [hunk 8]
```diff
@@ -1,18 +1,79 @@
...
+        cy.get('button').contains('Admin').click()
         cy.get('a').contains('PS1 Containment').click()
         cy.get('h3').contains('Containment').should('be.visible')
     })
 })
+
+const changeQuestionFields = (questionNo: number, newTitle: string, newSupportNotes: string, organization: string) => {
+    cy.getByDataTestid('edit-question-' + questionNo).click()
+    cy.getByDataTestid('question-title-' + questionNo).clear()
+    cy.getByDataTestid('question-title-' + questionNo).type(newTitle)
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+    cy.getByDataTestid('markdown-editor')
+        .shadow()
+        .within(() => {
+            cy.get('[id=editor]').replace(newSupportNotes)
+        })
```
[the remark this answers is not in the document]
petterwildhagen: No. I prefer keeping everything inside the test - not create extra abstraction layers outside the test.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+            const newSupportNotes = 'These are the new support notes!!!' + faker.random.words(2)
+            cy.getByDataTestid('organization-' + questionNo).then(function (t) {
+                let currentOrg = Cypress.$(t).text()
+                let organization = faker.random.arrayElement(Object.keys(Organization).filter(t => t !== currentOrg))
+
+                changeQuestionFields(questionNo, newTitle, newSupportNotes, organization)
+                cy.getByDataTestid('save-question').click()
+                cy.getByDataTestid('question-title-' + questionNo).should('have.text', newTitle)
+                cy.getByDataTestid('organization-' + questionNo).should('have.text', organization)
+                cy.getByDataTestid('question-title-' + questionNo)
```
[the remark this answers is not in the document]
petterwildhagen: Yes, I have changed my mind, I dont think we need these support classes,  I prefer to keep everything inside the test - not create extra abstraction layers

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+                    .parent()
+                    .within(() => {
+                        cy.contains(newSupportNotes).should('be.visible')
+                    })
+            })
+        })
+    })
+    it('Cancel Edit question - changes to title, support notes and organization are not saved', () => {
```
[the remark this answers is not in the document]
petterwildhagen: No, this will not work becuase the 'Navigate to a barrier' test would then not work, beforeEach is run for all tests.
beforeEach has to be used for what is common to all tests.
But parameterized the test so removed duplication. Fixed.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+                let currentOrg = Cypress.$(t).text()
+                let organization = faker.random.arrayElement(Object.keys(Organization).filter(t => t !== currentOrg))
+                changeQuestionFields(questionNo, newTitle, newSupportNotes, organization)
+                cy.getByDataTestid('cancel-edit-question').click()
+                cy.getByDataTestid('question-title-' + questionNo).should('not.have.text', newTitle)
+                cy.getByDataTestid('organization-' + questionNo).should('not.have.text', organization)
+                cy.getByDataTestid('question-title-' + questionNo)
+                    .next()
+                    .next()
+                    .should('not.have.text', newSupportNotes)
```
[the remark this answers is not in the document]
petterwildhagen: Well spotted. Fixed.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
+    })
+    it('Cancel Edit question - changes to title, support notes and organization are not saved', () => {
+        cy.get('[id^=question-]').then(questions => {
+            var questionsCount = Cypress.$(questions).length
+            const questionNo = faker.datatype.number({ min: 1, max: questionsCount })
+            const newTitle = 'This is the new title' + faker.random.words(2)
+            const newSupportNotes = 'These are the new support notes!!!' + faker.random.words(2)
+            cy.getByDataTestid('organization-' + questionNo).then(function (t) {
+                let currentOrg = Cypress.$(t).text()
+                let organization = faker.random.arrayElement(Object.keys(Organization).filter(t => t !== currentOrg))
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Deleted

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: You have to use the parent because the element I am looking for is on the same level as the support notes

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -1,18 +1,79 @@
...
-    })
-
-    it('Navigate to Admin page', () => {
         cy.get('button').contains('Admin').click()
         cy.contains('Project configuration: Questionnaire').should('be.visible')
     })
 
+    it('Edit question - change title, support notes and organization', () => {
+        cy.get('[id^=question-]').then(questions => {
+            var questionsCount = Cypress.$(questions).length
```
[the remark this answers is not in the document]
petterwildhagen: Yes, this is fine. Navigation is in other test that can be extended

achaikou (commented):
Don't have time today to check properly, did just a quick look, will after the tests are fixed. 
But in general I like it now :heart: 

As pageobjects location is important to you, then sure, I agree to move it out of support. I think team won't mind, but we'll discuss on the next meeting.

frontend/cypress/page_objects/admin_page.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    saveEditButton = () => {
+        return cy.getByDataTestid('save-question')
+    }
+
+    editQuestionButton = (questionNo: number) => {
+        return cy.getByDataTestid('edit-question-' + questionNo)
+    }
+
+    deleteAndInsertSupportNotes = (supportNotes: string) => {
```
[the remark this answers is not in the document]
petterwildhagen: Fixed.

achaikou (approved):
Looks good to me!
Markdown editor functionality can be moved to its own class by someone else later when it's reused first time.

Issue #534, opened 2021-08-17
BMT Teststrategy: Test automation


petterwildhagen:
1. Functionality (add/edit/delete actions etc)
2. Available functionality per role/per step

- [ ] #855

achaikou:
Note that role-first approach wouldn't make much sense for majority of the functionality as it would be the same for everyone.
So I think we better go with "What can be done by whom" instead of "Who can do what".

PR #663, opened 2021-09-22
Facilitator (other users) can (not) view/edit content on workshop summary tab

hnformentin:
Solves #534 

(1) The function `logInUser` I got from actions.ts. Not sure there is a better way to 'import' a function from cypress test.

(2) Having difficulties to test the disabler_box:

In the first test (Facilitator), none of these works, but maybe it is not necessary to test the disabler box anyway:
```

            const disabler_box = cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.getByDataTestid('disabler_box_div')
            })
            expect(disabler_box).to.have.attr('display', 'none')
            expect(disabler_box).to.have.css('display', 'none')
            disabler_box.should('have.css', 'display', 'none')
            disabler_box.should('be.hidden')


```

In addition, it does not raise an error with the following (which is the test run in the test for Participant, but for Facilitator should be `none`).

`            disabler_box.should('have.css', 'display', 'block')
`

The following also did not work:
```
            cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.get('fusion-markdown-editor').should('be.visible')
            })
```
giving the error:

```
AssertionError
Timed out retrying after 4000ms: expected '<fusion-markdown-editor>' to be 'visible'

This element <fusion-markdown-editor> is not visible because it has an effective width and height of: 0 x 0 pixels.
```

For the second test (Participant), the attempt below times out.

`
cy.getByDataTestid('workshop_summary_div').should('be.hidden')
`

but the following works:

```
            cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.get('fusion-markdown-editor').should('be.hidden')
            })

```

If not enough, the disabler_box could be tested, but I am not sure it is the good way because `disabler_box.should('have.css', 'display', 'block')` did not raised any error when testing with Facilitator logged in.

```
           const disabler_box = cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.getByDataTestid('disabler_box_div')
            })
            disabler_box.should('be.visible')
            disabler_box.should('have.css', 'display', 'block')

```

(3) In `workshop_assesssment.ts`, need to fix `WorkshopAssessmentTabs`, the current solution is unstable.

(4) Possibly won't need the data-testid="disabler_box_div", and if needed should be on a different commit.

### DIFF
--- a/frontend/cypress/integration/workshop_assessment_spec.ts
+++ b/frontend/cypress/integration/workshop_assessment_spec.ts
@@ -1,0 +1,58 @@
+import { EvaluationSeed } from '../support/testsetup/evaluation_seed'
+import { Progression, Role } from '../../src/api/models'
+import { WorkshopAssessmentTabs, WorkshopSummary } from '../page_objects/workshop_assessment'
+import { getUsers } from '../support/mock/external/users'
+import { MarkdownEditor, SaveIndicator } from '../page_objects/common'
+import { EvaluationPage } from '../page_objects/evaluation'
+import { SavingState } from '../../src/utils/Variables'
+
+describe('Workshop assessment', () => {
+    let seed: EvaluationSeed
+    const workshopAssessmentTabs = new WorkshopAssessmentTabs()
+    const workshopSummary = new WorkshopSummary()
+    const originalSummaryMessage = 'Chocolate for you!'
+    const editedSummaryMessage = 'Tulips for you! 🌷🌷🌷'
+    const progression = Progression.Workshop
+    const evaluationPage = new EvaluationPage()
+    const markdownEditor = new MarkdownEditor()
+
+    context('Workshop Summary', () => {
+        beforeEach(() => {
+            seed = new EvaluationSeed({
+                progression,
+                users: getUsers(4),
+                roles: [Role.Facilitator, Role.OrganizationLead, Role.Participant, Role.ReadOnly],
+            })
+            seed.addSummary({ summary: originalSummaryMessage, createdBy: seed.findParticipantByRole(Role.Facilitator) })
+            seed.plant()
+        })
+
+        it('FACILITATOR can edit and save notes', () => {
+            cy.visitProgression(progression, seed.evaluationId, seed.findParticipantByRole(Role.Facilitator).user)
+            workshopAssessmentTabs.workshopSummary().click()
+            markdownEditor.assertContent(originalSummaryMessage)
+            workshopSummary.assertEditorDisabler('hidden')
+            workshopSummary.body().within(() => {
+                markdownEditor.setContent(editedSummaryMessage)
+            })
+            new SaveIndicator().assertState(SavingState.Saved)
+
+            cy.testCacheAndDB(() => {
+                evaluationPage.progressionStepLink(progression).click()
+                workshopAssessmentTabs.workshopSummary().click()
+                markdownEditor.assertContent(editedSummaryMessage)
+            })
+        })
+
+        const roles = [Role.OrganizationLead, Role.Participant, Role.ReadOnly]
+
+        roles.forEach(role => {
+            it(`${role} can see Workshop Summary notes but not edit it`, () => {
+                cy.visitProgression(progression, seed.evaluationId, seed.findParticipantByRole(role).user)
+                workshopAssessmentTabs.workshopSummary().click()
+                markdownEditor.assertContent(originalSummaryMessage)
+                workshopSummary.assertEditorDisabler('visible')
+            })
+        })
+    })
+})
--- a/frontend/cypress/page_objects/common.ts
+++ b/frontend/cypress/page_objects/common.ts
@@ -1,3 +1,5 @@
+import { SavingState } from '../../src/utils/Variables'
+
 /* It's not clear yet if common components won't clash with each other */
 export class ConfirmationDialog {
     self = () => {
@@ -59,5 +61,29 @@ export class DropdownSelect {
     select = (selectInput: Cypress.Chainable, text: string) => {
         selectInput.click()
         cy.get('section').contains('span', text).click()
+    }
+}
+
+export class MarkdownEditor {
+    setContent = (content: string) => {
+        cy.get('fusion-markdown-editor')
+            .shadow()
+            .within(() => {
+                cy.get('[id=editor]').replace(content)
+            })
+    }
+
+    assertContent = (content: string) => {
+        cy.contains(content, { includeShadowDom: true }).should('be.visible')
+    }
+}
+
+export class SaveIndicator {
+    assertState = (state: SavingState) => {
+        if (state === SavingState.Saved) {
+            cy.getByDataTestid('save_indicator', 10000).should('have.text', 'Saved')
+        } else {
+            throw new Error('SavingState Saving, NotSaved and None not implemented for tests.')
+        }
     }
 }
--- a/frontend/cypress/page_objects/workshop_assessment.ts
+++ b/frontend/cypress/page_objects/workshop_assessment.ts
@@ -1,0 +1,23 @@
+export class WorkshopAssessmentTabs {
+    workshopSummary = () => {
+        return cy.getByDataTestid('workshop_assessment_tablist').contains('Workshop Summary')
+    }
+}
+
+export class WorkshopSummary {
+    body = () => {
+        return cy.getByDataTestid('workshop_summary_div')
+    }
+
+    assertEditorDisabler = (boxStatus: 'hidden' | 'visible') => {
+        if (boxStatus === 'visible') {
+            this.body().within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'block')
+            })
+        } else if (boxStatus === 'hidden') {
+            this.body().within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'none')
+            })
+        }
+    }
+}
--- a/frontend/cypress/support/commands/visit.ts
+++ b/frontend/cypress/support/commands/visit.ts
@@ -1,5 +1,8 @@
 import { User } from '../mock/external/users'
 import { getFusionProjectData, findFusionProjectByID } from '../mock/external/projects'
+import { Evaluation, Progression } from '../../../src/api/models'
+import { EvaluationPage } from '../../page_objects/evaluation'
+
 function setupEnvironment() {
     cy.interceptExternal()
     cy.viewport(1800, 1000) //until we decide to test on various resolutions
@@ -66,6 +69,12 @@ Cypress.Commands.add('visitEvaluation', (evaluationId: string, user: User, fusio
     waitForEvaluationPageLoad()
 })
 
+Cypress.Commands.add('visitProgression', (progression: Progression, evaluationId: string, user: User, fusionProjectId: string = '123') => {
+    const evaluationPage = new EvaluationPage()
+    cy.visitEvaluation(evaluationId, user, fusionProjectId)
+    evaluationPage.progressionStepLink(progression).click()
+})
+
 Cypress.Commands.add('reloadBmt', (user: User, fusionProjectId: string = '123') => {
     /**
      * Situation with reload is quite weird. At the moment all problems are
@@ -105,6 +114,12 @@ declare global {
              * @example cy.visitEvaluation(evaluationId, extHire)
              */
             visitEvaluation(evaluationId: string, user: User, fusionProjectId?: string): Chainable<void>
+
+            /**
+             * Visit evaluation at a specific progression
+             * @example cy.visitProgression(progression, evaluation, user, project)
+             */
+            visitProgression(progression: Progression, evaluationId: string, user: User, fusionProjectId?: string): Chainable<void>
 
             /**
              * Reloads the page and attempts to deal with external dependencies issues
--- a/frontend/src/components/Disabler.tsx
+++ b/frontend/src/components/Disabler.tsx
@@ -21,6 +21,7 @@ const Disabler = ({ disable, children }: PropsWithChildren<Props>) => {
                         height: '100%',
                         width: '100%',
                     }}
+                    data-testid="disabler_box_div"
                 />
                 {children}
             </div>
--- a/frontend/cypress/integration/actions_spec.ts
+++ b/frontend/cypress/integration/actions_spec.ts
@@ -32,17 +32,6 @@ describe('Actions management', () => {
         evaluationPage.progressionStepLink(progression).click()
         return user
     }
-    const logInUser = (role: Role, progression: Progression = getRandomProgressionWorkshopOrFollowUp()): User => {
-        const participant = seed.participants.find(x => {
-            return x.role === role
-        })
-        if (participant === undefined) {
-            throw 'No user with role ' + role + ' found'
-        }
-        cy.visitEvaluation(seed.evaluationId, participant.user)
-        evaluationPage.progressionStepLink(progression).click()
-        return participant.user
-    }
 
     beforeEach(() => {
         seed = createSeedWithActions(users, roles, { completed: false })
@@ -69,7 +58,7 @@ describe('Actions management', () => {
         ]
         roles.forEach(r => {
             it(`${r.role} can create action = ${r.canCreateAction}`, () => {
-                logInUser(r.role)
+                cy.visitProgression(getRandomProgressionWorkshopOrFollowUp(), seed.evaluationId, seed.findParticipantByRole(r.role).user)
                 let action = createActionTestData(users)
                 r.canCreateAction
                     ? actionsGrid.addActionButton(action.questionOrder).should('be.enabled')
@@ -80,7 +69,7 @@ describe('Actions management', () => {
         const randomRole = faker.random.arrayElement([Role.Facilitator, Role.Participant, Role.OrganizationLead])
 
         it(`Action creation by ${randomRole} and verification that action was created`, () => {
-            logInUser(randomRole)
+            cy.visitProgression(getRandomProgressionWorkshopOrFollowUp(), seed.evaluationId, seed.findParticipantByRole(randomRole).user)
             let action = createActionTestData(users)
             actionsGrid.addActionButton(action.questionOrder).click()
             createActionDialog.titleInput().type(action.title)
@@ -121,7 +110,8 @@ describe('Actions management', () => {
         let newNotes: Note[]
 
         it(`Editing action by ${randomRole} - assign action, add notes and verify notes were added`, () => {
-            let user = logInUser(randomRole)
+            let user = seed.findParticipantByRole(randomRole).user
+            cy.visitProgression(getRandomProgressionWorkshopOrFollowUp(), seed.evaluationId, user)
             ;({ actionNotes, action } = findActionWithNotes(seed))
             ;({ updatedAction, newNotes } = createEditTestData(seed, user, action))
 
@@ -169,7 +159,7 @@ describe('Actions management', () => {
         ]
         roles.forEach(r => {
             it(`${r.role} delete button exists = ${r.deleteButtonExists} can delete action = ${r.canDeleteAction}`, () => {
-                logInUser(r.role)
+                cy.visitProgression(getRandomProgressionWorkshopOrFollowUp(), seed.evaluationId, seed.findParticipantByRole(r.role).user)
                 const { actionToDelete, actionToStay } = getActionToDeleteActionToStay()
                 if (!r.deleteButtonExists) {
                     actionsGrid.deleteActionButton(actionToDelete.id).should('not.exist')
@@ -196,7 +186,11 @@ describe('Actions management', () => {
         }
 
         it(`Cancel delete action by Facilitator, then verify action was not deleted`, () => {
-            logInUser(Role.Facilitator)
+            cy.visitProgression(
+                getRandomProgressionWorkshopOrFollowUp(),
+                seed.evaluationId,
+                seed.findParticipantByRole(Role.Facilitator).user
+            )
             const { actionToDelete, actionToStay } = getActionToDeleteActionToStay()
             actionsGrid.deleteActionButton(actionToDelete.id).click()
             confirmationDialog.noButton().click()
@@ -209,7 +203,11 @@ describe('Actions management', () => {
         })
         context('Error handling when facilitator has two browser windows open', () => {
             it('Deleted action in one window cannot be deleted in other window', () => {
-                logInUser(Role.Facilitator)
+                cy.visitProgression(
+                    getRandomProgressionWorkshopOrFollowUp(),
+                    seed.evaluationId,
+                    seed.findParticipantByRole(Role.Facilitator).user
+                )
                 const { actionToDelete, actionToStay } = getActionToDeleteActionToStay()
                 cy.gql(DELETE_ACTION, {
                     variables: {
@@ -289,7 +287,8 @@ describe('Actions management', () => {
         const roleThatCanComplete = faker.random.arrayElement([Role.Facilitator, Role.Participant, Role.OrganizationLead])
 
         it(`Action can be completed by ${roleThatCanComplete} and must have a reason`, () => {
-            let user = logInUser(roleThatCanComplete)
+            let user = seed.findParticipantByRole(roleThatCanComplete).user
+            cy.visitProgression(getRandomProgressionWorkshopOrFollowUp(), seed.evaluationId, user)
             ;({ actionNotes, action } = findActionWithNotes(seed))
             ;({ updatedAction, newNote } = createCompleteAction(user, action))
 
@@ -317,7 +316,11 @@ describe('Actions management', () => {
         })
 
         it(`Completing Action can be cancelled by ${roleThatCanComplete}`, () => {
-            let user = logInUser(roleThatCanComplete)
+            cy.visitProgression(
+                getRandomProgressionWorkshopOrFollowUp(),
+                seed.evaluationId,
+                seed.findParticipantByRole(roleThatCanComplete).user
+            )
             ;({ actionNotes, action } = findActionWithNotes(seed))
 
             actionsGrid.actionLink(action.questionOrder, action.title).click()
--- a/frontend/cypress/page_objects/action.ts
+++ b/frontend/cypress/page_objects/action.ts
@@ -1,8 +1,9 @@
-import { DropdownSelect, SideSheet } from '../page_objects/common'
+import { DropdownSelect, SideSheet, SaveIndicator } from '../page_objects/common'
 import { Action, Note } from '../support/testsetup/mocks'
 import { FUSION_DATE_LOCALE } from '../support/helpers/helpers'
 import { Priority, Question } from '../../src/api/models'
 import { barrierToString, organizationToString } from '../../src/utils/EnumToString'
+import { SavingState } from '../../src/utils/Variables'
 
 /**
  * List of actions under every question
@@ -147,7 +148,7 @@ export class EditActionDialog extends ActionDialog {
             .parent()
             .parent()
             .within(() => {
-                cy.getByDataTestid('save_indicator', 10000).should('have.text', 'Saved')
+                new SaveIndicator().assertState(SavingState.Saved)
             })
     }
 
--- a/frontend/cypress/support/testsetup/evaluation_seed.ts
+++ b/frontend/cypress/support/testsetup/evaluation_seed.ts
@@ -136,6 +136,16 @@ export class EvaluationSeed {
         return question.id
     }
 
+    findParticipantByRole(desiredRole: Role) {
+        const participant = this.participants.find(x => {
+            return x.role === desiredRole
+        })
+        if (participant === undefined) {
+            throw 'No user with role ' + desiredRole + ' found'
+        }
+        return participant
+    }
+
     plant() {
         const facilitator = this.participants.find(e => e.role === Role.Facilitator)
         if (facilitator === undefined) {
--- a/frontend/src/views/Evaluation/Workshop/WorkshopSummaryWithApi.tsx
+++ b/frontend/src/views/Evaluation/Workshop/WorkshopSummaryWithApi.tsx
@@ -47,7 +47,7 @@ const WorkshopSummaryWithApi = ({ evaluation, disable }: React.PropsWithChildren
     }
 
     return (
-        <div style={{ padding: '30px' }}>
+        <div style={{ padding: '30px' }} data-testid="workshop_summary_div">
             <WorkshopSummary localSummary={localSummary} onChange={onChange} savingState={savingState} disable={disable} />
         </div>
     )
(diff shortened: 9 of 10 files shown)


frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,60 @@
...
+                cy.getByDataTestid('save_indicator').should('be.visible')
+            })
+        })
+
+        it('ANY PARTICIPANT can not edit notes', () => {
+            logInUser(Role.Participant, Progression.Workshop)
+            workshopAssessmentTabs.workshopSummary().click()
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.get('fusion-markdown-editor').should('be.hidden')
```
achaikou, line 52: > (2)

Yep, you are correct. If "be.visible" doesn't work in the other test, we can't really use it here. Because if the field suddenly becomes enabled, this div will still stay hidden, so we won't discover the failure. (I suspect now that checking for "visible" elements doesn't work if they are just divs without taking any actual space in the code).
Thus I totally agree with all your suspicions!

> but I am not sure it is the good way because disabler_box.should('have.css', 'display', 'block') did not raised any error when testing with Facilitator logged in.

Very good testing mindset! :nerd_face: :clap: :heart: 

---
I tried to play with these hidden/visible and unfortunately it didn't work. Cypress has a lot of internal logic when clicking elements to discover if they are visible or not. Unfortunately this logic is not accessible for us directly.
So my suggestions are

1. The best option probably would be to disable the editor a different way. Maybe it's actually a good idea to talk to Nina first: now we know that there exists component MarkdownViewer. So maybe it would be better user experience if we didn't cover the MarkdownEditor with Disabler box, but instead rendered MarkdownViewer for users who are not supposed to edit the content?
If Nina approves, then first task might be just to exchange Editor with Viewer - and problem solved as I suspect for Participant you would have` fusion-markdown-viewer `instead of `fusion-markdown-editor` in the code.

2. Try to kinda bypass cypress and use its internal logic by attempting to type into this editor - and then (what is not recommended usually) intercept cypress failure 
```
cy.on('fail', (err, runnable) => {
  if (err.message.includes('is being covered by another element')) {
        if (user.id == "222") { // whatever good check: if there is a failure for expected user - return
              return
        }
  }
  throw err
})
```
The only problem with this solution is that if Participant somehow can edit the field, this message won't be thrown, and we won't catch the error. To take care of that we simply would need to check if the value exists after typing: for Facilitator we check that typed value is still visible and that for Participant it is not.
 
From the point of stability and user-perspective I like this solution more than the next, even though it is a bit complicated and uses on.fail (though I think the use case here is valid)
 
3. Fix the "block" thing. I am not very fond of it, as it is bypassing the actual thing we want to test, but it is easy and straightforward.

I think the problem is that you do not use cypress correctly. You probably fall under one of the most irritating things in cypress: [within returns](https://docs.cypress.io/api/commands/within#Yields) **outer** element, not **inner**. That's why in many cases I couldn't just put locator to support files and to do "clickThisThingThatIsInsideWithinCallBecauseIJustCantRetunTheThingItself". Apparently there are workarounds, but still :unamused: 
So I think that your code was actually checking `workshop_summary_div`, not the `disabler_box_div` (you can hover over it in log and see that the full field is covered)

So almost-your-code works correctly for me.
```
            cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'none')
```
Passes for Facilitator, fails for Participant.

```
            cy.getByDataTestid('workshop_summary_div').within(() => {
                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'block')
```
Fails for Facilitator, passes for Participant.   
 
 ----
Piece of my mind given, see for yourself what works best! :smile:

frontend/src/components/Disabler.tsx
```diff
@@ -21,6 +21,7 @@ const Disabler = ({ disable, children }: PropsWithChildren<Props>) => {
                         height: '100%',
                         width: '100%',
                     }}
+                    data-testid="disabler_box_div"
```
achaikou, line 24: > (4) Possibly won't need the data-testid="disabler_box_div", and if needed should be on a different commit.

Would depend on the solution chosen, of course, but not sure what would be the point of different commit? Property has no impact on source code, this is purely part of the test. If I saw it as a separate commit I would be on opposite surprised: why? what purpose does this change hold?

achaikou (commented):
Comments related to your questions only :slightly_smiling_face:

frontend/cypress/support/workshop_assessment.ts
```diff
@@ -0,0 +1,23 @@
+export default class WorkshopAssessmentTabs {
+    workshopSummary = () => {
+        // return cy.get('#fixed-tablist').contains('Workshop Summary')
+        // fix this solution
+        return cy.get('#tabs-1-tab-2')
+    }
```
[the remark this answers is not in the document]
hnformentin: Good suggestion. I think that would not be a problem to have only `cy.contains("Workshop Summary")`, but I thought it would be cleaner to include a data-testid for `workshop_assessment_tablist`.

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: @denektenina and me agreed that using `MarkdownViewer` would be less intuitive to the user: while MarkdownEditor 
 + Disabler box let clear that it is an editable field (maybe not just accessible for that specific user), only MarkdownViewer could give the impression that the text was not editable at all.
I found the second option more confusing than the third one for future developers, therefore, I opted to the third option.

So rich to me to have options written so clearly. Thanks :-)

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
+import { EvaluationSeed } from '../support/evaluation_seed'
+import { Progression, Role } from '../../src/api/models'
+import WorkshopAssessmentTabs from '../support/workshop_assessment'
+import { EvaluationPage } from '../page_objects/evaluation'
+import { getUsers, User } from '../support/mock/external/users'
```
achaikou, line 5: Unused import

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        })
+        it('FACILITATOR can edit and save notes', () => {
+            cy.visitProgression(Progression.Workshop, seed.evaluationId, seed.findParticipantByRole(Role.Facilitator).user)
+            workshopAssessmentTabs.workshopSummary().click()
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.get('fusion-markdown-editor')
+                    .shadow()
+                    .within(() => {
+                        cy.get('[id=editor]').replace('Tulips for you! 🌷🌷🌷')
+                    })
```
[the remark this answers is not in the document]
petterwildhagen: Or create cypress/support/markdown editor.ts and put it in there? 
Page_objects (I am not sure we need them) should contain elements related to specific 
pages.  Without page_objects you would write 
new MarkdownEditor().setContent(supportNotes)
directly in the (in this example) admin_page tests

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('be.visible')
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'none')
+            })
```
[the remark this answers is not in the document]
petterwildhagen: Exactly. This is the main argument against page object files. Re use outside the test is unlikely.
If the data-testid attributes are descriptive you can write cy.getByDataTestid(descriptive_data_ testis).click() in the test itself instead of putting it in a page_object file. This removes one abstraction layer (reduces complexity)

frontend/cypress/support/workshop_assessment.ts
```diff
@@ -0,0 +1,21 @@
+export default class WorkshopAssessmentTabs {
```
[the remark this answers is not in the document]
hnformentin: Good observation.

frontend/cypress/support/workshop_assessment.ts
```diff
@@ -0,0 +1,21 @@
...
+    assertSaved = () => {
+        this.body()
+            .parent()
+            .parent()
+            .parent()
+            .within(() => {
+                cy.getByDataTestid('save_indicator', 10000).should('have.text', 'Saved')
+            })
+    }
+}
```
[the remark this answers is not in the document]
hnformentin: My bad. In the beginning I thought I would use as save indicator, but did not.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                    .shadow()
+                    .within(() => {
+                        cy.get('[id=editor]').replace('Tulips for you! 🌷🌷🌷')
+                    })
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('be.visible')
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'none')
```
petterwildhagen, line 35: Replace with cy.getByDataTestid('disabler_box_div').should('not.be.visible') ( less tied to implememtation)

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+                cy.get('fusion-markdown-editor').should('be.hidden')
+            })
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('not.exist')
+            })
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'block')
+            })
```
petterwildhagen, line 53: replace with cy.getByDataTestid('disabler_box_div').should('be.visible') (less tied to impelemtation)

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('be.visible')
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'none')
+            })
+        })
+
+        it('ANY PARTICIPANT can not edit notes', () => {
```
petterwildhagen, line 39: Change to 'Any PARTICIPANT can only view notes '
and then check that notes display (extension) in addition to edit not being possible?

frontend/cypress/integration/workshop_assessment_spec.ts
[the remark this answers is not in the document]
hnformentin: If I understood well, @achaikou and @petterwildhagen arguments are against each other.
Both considers readability and layers of abstraction in a different way, but in a different way, if I got it properly.

Did you discussed something around this on Friday's meeting?

If not, I would suggest an adaptation that is `workshopSummary.assertEditorDisablerDisabled(True)` or `workshopSummary.assertEditorDisablerDisabled(False)`.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.get('fusion-markdown-editor')
+                    .shadow()
+                    .within(() => {
+                        cy.get('[id=editor]').replace('Tulips for you! 🌷🌷🌷')
+                    })
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('be.visible')
+            })
```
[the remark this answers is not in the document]
hnformentin: Makes sense.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            cy.visitProgression(Progression.Workshop, seed.evaluationId, seed.findParticipantByRole(Role.Participant).user)
+            workshopAssessmentTabs.workshopSummary().click()
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.get('fusion-markdown-editor').should('be.hidden')
+            })
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('save_indicator').should('not.exist')
+            })
```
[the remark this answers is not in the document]
hnformentin: Makes sense.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    const workshopAssessmentTabs = new WorkshopAssessmentTabs()
+
+    context('Workshop Summary', () => {
+        beforeEach(() => {
+            seed = new EvaluationSeed({
+                progression: Progression.Workshop,
+                users: getUsers(2),
+                roles: [Role.Facilitator, Role.Participant],
+            })
+            seed.plant()
```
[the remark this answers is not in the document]
hnformentin: Good idea. I see the point of creating an evaluation with a summary already there. 
In regards to the user roles, I thought testing all the roles would be heavy, that is why I only tested Facilitator and Participant. Do you think I would need 
(a) to create an evaluation having users with each role and them test each of them
(b) have an evaluation with two participants (one Facilitator, one random user) and test only this one?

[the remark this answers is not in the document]
achaikou: I think you need option a) as I think we agreed that Roles are our main "parameter" in bmt and we want it tested thoroughly.

It won't really be heavy or long. You just log with 4 users and expect 3 of them to have the summary field disabled, so don't do anything else. And there is just 1 user (Facilitator) who can perform the operation itself.

Related to the comment 
https://github.com/equinor/fusion-bmt/pull/663#discussion_r718688729

frontend/cypress/integration/workshop_assessment_spec.ts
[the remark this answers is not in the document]
achaikou: I think that now we agreed that per our guidelines we want this additional abstraction level.
So (1) now is preferable to (2)

frontend/cypress/support/workshop_assessment.ts
```diff
@@ -0,0 +1,21 @@
+export default class WorkshopAssessmentTabs {
+    workshopSummary = () => {
+        return cy.getByDataTestid('workshop_assessment_tablist').contains('Workshop Summary')
+    }
+}
+
```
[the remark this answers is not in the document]
hnformentin: Thanks, I will update!

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            })
+        })
+
+        it('ANY PARTICIPANT can not edit notes', () => {
+            cy.visitProgression(Progression.Workshop, seed.evaluationId, seed.findParticipantByRole(Role.Participant).user)
+            workshopAssessmentTabs.workshopSummary().click()
+
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.get('fusion-markdown-editor').should('be.hidden')
+            })
```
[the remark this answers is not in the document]
hnformentin: Thanks. I did not noticed that. But after your comment, I tested the same code on the test above and it passes, what reinforce this is not a good test.

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: Why? This test is now checking that the disabler_box is there, not that the user can see the content. But there is a suggestion up in the review to refactor deeper the test, in case we also assert that the user can read a workshop summary, then this makes sense...

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: I am not sure...I think we are checking though the chain of elements. 
By implementing your suggestion, if we have a disabler_box outside the workshop_summary_div, the test would pass but it actually should not. So I think it is better to keep like that.

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: Same comment as for your suggestion below.

frontend/cypress/page_objects/workshop_assessment.ts
```diff
@@ -0,0 +1,17 @@
...
+        return cy.getByDataTestid('workshop_assessment_tablist').contains('Workshop Summary')
+    }
+    // How to make it within workshopSummary
+    assertEditorDisablerDisabled = (disabled: boolean) => {
+        // something wrong and it is not entering the if
+        if (disabled) {
+            cy.getByDataTestid('workshop_summary_div').within(() => {
+                cy.getByDataTestid('disabler_box_div').should('have.css', 'display', 'block')
+            })
+        }
```
hnformentin, line 12: Why is it not entering the if statement even when I pass `disabled true`?

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                })
+                // by
+                workshopAssessmentTabs.assertEditorDisablerDisabled(true)
+                // check content
+                // solution 1
+                // cy.contains('Chocolate for you!')
+                // solution 2, getContent returns void?
+                // cy.getByDataTestid('workshop_summary_div').within(() => {
+                //     new MarkdownEditor().getContent().should('contain', 'Chocolate for you!')
+                // })
```
hnformentin, line 62: Why none of the solutions is able to identify 'Chocolate for you!', even if it is there when I inspect?

frontend/cypress/page_objects/workshop_assessment.ts
achaikou: Oh, it does enter it

image


Problem is that you forgot `else` below :) 
So if first does `disabled? check "block"` and then anyway checks for "none", regardless of disabled value.

frontend/cypress/integration/workshop_assessment_spec.ts
achaikou: Remember the bad "shadow dom root" thing? It screws us here too!
Content is inside shadow, so it is not found, unless you pass [includeShadowDom](https://docs.cypress.io/api/commands/contains#Arguments) option.

`cy.contains('Chocolate for you!', {includeShadowDom: true}).should("be.visible")`

frontend/cypress/page_objects/workshop_assessment.ts
hnformentin: Excellent, I was thinking `cy.get...` as a return statement instead of an assertion...thanks!

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                cy.visitProgression(Progression.Workshop, seed.evaluationId, seed.findParticipantByRole(Role.Facilitator).user)
+                workshopAssessmentTabs.workshopSummary().click()
+                cy.contains(originalSummaryMessage, { includeShadowDom: true }).should('be.visible')
+                workshopSummary.assertEditorDisablerDisabled(false)
+                cy.getByDataTestid('workshop_summary_div').within(() => {
+                    new MarkdownEditor().setContent(editedSummaryMessage)
+                })
+                cy.contains(editedSummaryMessage)
+            })
+            cy.getByDataTestid('workshop_summary_div').within(() => {
```
achaikou, line 36: I still vote into moving all these things to support files 
`cy.getByDataTestid('workshop_summary_div')` could be `WorkshopSummary.body`, for example (yes, I am promoting my naming :sweat_smile: ) and the call itself might be `WorkshopSummary.assertSave(state)`. 
But I would honestly prefer to create a SaveIndicator component (similar to MarkdownEditor) and have `SaveIndicator.assertState(state)` because it is used more than 5 times in frontend code, so will be reused in tests as well.

frontend/cypress/page_objects/workshop_assessment.ts
```diff
@@ -0,0 +1,19 @@
+export class WorkshopAssessmentTabs {
+    workshopSummary = () => {
+        return cy.getByDataTestid('workshop_assessment_tablist').contains('Workshop Summary')
+    }
+}
+
+export class WorkshopSummary {
+    assertEditorDisablerDisabled = (disabled: boolean) => {
```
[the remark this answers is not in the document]
petterwildhagen: You could define an enum (which might be an overkill) or just check in the function itself. I would favour the latter. In the function itself you could  create a switch statement and throw an exception if the argument does not match.

frontend/cypress/integration/workshop_assessment_spec.ts
petterwildhagen: We sort of established this as a best practice on friday, so agreed , we should go with this.

frontend/cypress/page_objects/workshop_assessment.ts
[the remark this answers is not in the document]
achaikou: 1. I am not good enough with typescript to check what exactly your code does, but this syntax is for [assigning default value](https://www.typescripttutorial.net/typescript-tutorial/typescript-default-parameters/), in your case - just union of strings `'disabled' || 'visible'` probably?
2. Apparently the correct [way is to do](https://stackoverflow.com/a/35312363) it as we did in here https://github.com/equinor/fusion-bmt/blob/d55783ad38a56a88416377b72ec472395428e25b/frontend/cypress/page_objects/evaluation.ts#L5 So syntax is a bit different as here you restrict type, not value
3. I am not sure I am fond of "disabler disabled" still :D After all, if display = none, the element simply takes no space on the page, not that it does but is just disabled... Maybe visible vs hidden (as in Cypress) is nicer that visible vs disabled.
4. To the following: do we even have to check explicitly for display block property? Or can we simply write
```
cy.getByDataTestid('disabler_box_div').should('be.hidden') //if hidden
cy.getByDataTestid('disabler_box_div').should('be.visible') //if visible
```
Current version with block/none is ok too (so don't change unless you want to), but maybe this can help a bit with naming, idk.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                cy.getByDataTestid('save_indicator').should('be.visible')
```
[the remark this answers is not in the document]
hnformentin: I will accept the suggestion of `cy.getByDataTestid('save_indicator').should('have.text', 'Saved')` without checking further details for being faster on addressing the comments, but I plan to come back to it later...

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,52 @@
...
+            })
+        })
+
+        const roles = [Role.OrganizationLead, Role.Participant, Role.ReadOnly]
+
+        roles.forEach(role => {
+            it(`${role} can see Workshop Summary notes but not edit it`, () => {
+                cy.visitProgression(Progression.Workshop, seed.evaluationId, seed.findParticipantByRole(role).user)
+                workshopAssessmentTabs.workshopSummary().click()
+                cy.contains(originalSummaryMessage, { includeShadowDom: true }).should('be.visible')
```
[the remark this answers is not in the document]
hnformentin: With our discussion it was clear that this is a good idea :-)

frontend/cypress/page_objects/common.ts
```diff
@@ -61,3 +61,25 @@ export class DropdownSelect {
...
+    }
+}
+
+export class SaveIndicator {
+    assertState = (state: string) => {
+        if (state === 'saved') {
+            cy.getByDataTestid('save_indicator').should('have.text', 'Saved')
+        }
+    }
+}
```
hnformentin, line 85: Not happy with the way I implemented here because if anything else comes as `state`, it will simply not do anything. Not sure if an error message would be the correct thing...

frontend/cypress/page_objects/workshop_assessment.ts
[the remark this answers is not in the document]
hnformentin: `hidden` or `visible` is a good suggestion. Thanks.

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -23,28 +26,29 @@ describe('Workshop assessment', () => {
         })
 
         it('FACILITATOR can edit and save notes', () => {
+            cy.visitProgression(progression, seed.evaluationId, seed.findParticipantByRole(Role.Facilitator).user)
+            workshopAssessmentTabs.workshopSummary().click()
+            new MarkdownEditor().assertContent(originalSummaryMessage)
```
achaikou, line 31: create a markdownEditor variable similar to workshopSummary? Using new every time is a bit weird

frontend/cypress/page_objects/common.ts
achaikou: 1.
You can also restrict it the same way you did with visible/disabled!

Or you can require frontend enum here SavingState
https://github.com/equinor/fusion-bmt/blob/2732488627acf13f819b1cdaeb8561bd625f8984/frontend/src/utils/Variables.ts#L10
I believe it is ok to reuse these values from frontend, as it is not test implementation, but simply utility type definition

2. It would also be good if you refactor existing code to use the same solution
aka exchange
https://github.com/equinor/fusion-bmt/blob/2732488627acf13f819b1cdaeb8561bd625f8984/frontend/cypress/page_objects/action.ts#L150
with
`new SaveIndicator().assertState("saved")`

frontend/cypress/integration/workshop_assessment_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            seed.addSummary({ summary: originalSummaryMessage, createdBy: seed.findParticipantByRole(Role.Facilitator) })
+            seed.plant()
+        })
+
+        it('FACILITATOR can edit and save notes', () => {
+            cy.visitProgression(progression, seed.evaluationId, seed.findParticipantByRole(Role.Facilitator).user)
+            workshopAssessmentTabs.workshopSummary().click()
+            new MarkdownEditor().assertContent(originalSummaryMessage)
+            workshopSummary.assertEditorDisabler('visible')
+            cy.getByDataTestid('workshop_summary_div').within(() => {
```
achaikou, line 33: Can you move this one `cy.getByDataTestid('workshop_summary_div')` to the adminPage locators as well (as `WorkshopSummary.body()` or something)? As Petter mentioned in another comment, this is what we already agreed upon

frontend/cypress/page_objects/workshop_assessment.ts
```diff
@@ -0,0 +1,19 @@
+export class WorkshopAssessmentTabs {
+    workshopSummary = () => {
+        return cy.getByDataTestid('workshop_assessment_tablist').contains('Workshop Summary')
+    }
+}
+
+export class WorkshopSummary {
+    assertEditorDisabler = (boxStatus: 'hidden' | 'visible') => {
+        if (boxStatus === 'hidden') {
```
achaikou, line 9: I think you have the opposite values here :)
assertEditorDisabler - visible? - display block
assertEditorDisabler - hidden? display none

for now test has to say `workshopSummary.assertEditorDisabler('visible')` for facilitator and that makes no sense

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: I think I understand now and it is implemented. Thanks.

frontend/cypress/integration/workshop_assessment_spec.ts
hnformentin: It worked. Thanks!

frontend/cypress/support/commands/visit.ts
```diff
@@ -66,6 +69,12 @@ Cypress.Commands.add('visitEvaluation', (evaluationId: string, user: User, fusio
     waitForEvaluationPageLoad()
 })
 
+Cypress.Commands.add('visitProgression', (progression: Progression, evaluationId: string, user: User, fusionProjectId: string = '123') => {
```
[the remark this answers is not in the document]
hnformentin: There are three additional functions in visit.ts with `fusionProjectId: string = '123'`.
Do you recommend to remove then also or only for `visitProgression`?

hnformentin:
Analysing what happens in the individual commits locally: each of the four passes all the tests.

- 7a837d2: Add cy.visitProgression, seed.findParticipantByRole 
Screenshot 2021-10-12 at 13 31 14
- 359d1e6: Refactor logInUser from actions_spec to cy.visitProgression
Screenshot 2021-10-12 at 13 19 32

- 6d467bc: Test facilitator (users) can (not) edit content on workshop summary tab
Screenshot 2021-10-12 at 13 40 53

- 4406208: Refactor actions.ts to use SaveIndicator from common
Screenshot 2021-10-12 at 14 30 03

ErlendHaa:
I like that you run tests for individual commits @hnformentin. Kodus! Here is a pro-tip for you:
```
git rebase -i --exec="cd frontend && npm run cyrun" HEAD~4
```

git will iterate through the last 4 commits and for each of them run `cd frontend && npm run cyrun`. (cd because it --exec is always executed from the repo root, regardless of the directory you run from).

As a bonus, if the commands fail (i.e. the cypress test fails) git will pause the rebase at that commit. Which means you can fix whatever is wrong and test it manually. Then commit using `--amend`, and run `git rebase --continue`

hnformentin:
Now, analysing why the pipeline failed in the second commit:

- Error in Cypress_E2E_Tests 1

```
Actions management
    Creating and editing actions
      1) "before each" hook for "FACILITATOR can create action = true"


  0 passing (3s)
  1 failing

  1) Actions management
       Creating and editing actions
         "before each" hook for "FACILITATOR can create action = true":
     CypressError: `cy.request()` failed trying to load:

http://localhost:5000/graphql

We attempted to make an http request to this URL but the request failed without a response.

We received this error at the network level:

  > Error: connect ECONNREFUSED 127.0.0.1:5000

-----------------------------------------------------------

The request we sent was:

Method: POST
URL: http://localhost:5000/graphql

-----------------------------------------------------------

Common situations why this would fail:
  - you don't have internet access
  - you forgot to run / boot your web server
  - your web server isn't accessible
  - you have weird network configuration settings on your computer

https://on.cypress.io/request

```

ErlendHaa:
That's the output you posted [above](https://github.com/equinor/fusion-bmt/pull/663#issuecomment-940996999) right? Seams to me to be the issue where the the tests starts running before the backend is ready. But maybe @achaikou or @petterwildhagen can back me up on that one?

PR #668, opened 2021-09-22
Add / remove project category label to/from question template

denektenina:
With this pull request it will be possible to add and remove a project category label to/from a question template (i.e. add/ remove question template from project category). The view looks like this:

image

By clicking the menu element next to the Barrier name, all questions in the barrier are made editable. It is now possible to select or deselect elements from the dropdown. Every selection triggers one API-call, either to add or remove.

**Features of this PR:**
* Frontend for adding and deleting as described above
* GraphQL mutations for adding and deleting
* Save-indicator on each question when saving
* Error message if saving fails
* Caching of results
* I also made a helper for some code that was used in several components, and used the helper in these "old" components as well

**Known potential problems**
* If an error occurs when saving or deleting a project category, an error message will appear, but there will also be an inconsistency in the select-dropdown and the actual tagged projects (shown in the amount of chips). For instance: If your question template has two categories, it will have two category-chips. If you want to remove one of the categories, and you deselect the category in the dropdown, but the call fails, the deselected category will still be deselected in the dropdown, but the number of chips will still be two chips, since no category has been deleted in the DB. This also happens when adding fails. _This is not fixed because it's because of how the Multiple Select handles its data, and I didn't want to spend time on something that might not be an issue. In my opinion, it's not important, since the inconsistency just highlights that the save failed, and the chips (and the text) reflect if the category really were added / deleted. When the edit-mode is closed, the inconsistency will no longer exist._

Illustration of problem:
image

Edit: Changed so that a "Not Saved"-status makes it even more clear that the change in the dropdown has not reached the DB:

image

### DIFF
--- a/frontend/src/views/Admin/BarrierMenu.tsx
+++ b/frontend/src/views/Admin/BarrierMenu.tsx
@@ -1,0 +1,30 @@
+import React, { RefObject } from 'react'
+import { Icon, Menu, Typography } from '@equinor/eds-core-react'
+import { add_circle_outlined, close_circle_outlined } from '@equinor/eds-icons'
+
+interface Props {
+    isOpen: boolean
+    anchorRef: RefObject<HTMLButtonElement>
+    closeMenu: () => void
+    setIsInAddCategoryMode: (inEditmode: boolean) => void
+    isInAddCategoryMode: boolean
+}
+
+const BarrierMenu = ({ isOpen, anchorRef, closeMenu, setIsInAddCategoryMode, isInAddCategoryMode }: Props) => {
+    return (
+        <Menu id="menu-complex" open={isOpen} anchorEl={anchorRef.current} onClose={closeMenu} placement={'bottom'}>
+            <Menu.Item
+                onClick={() => {
+                    setIsInAddCategoryMode(!isInAddCategoryMode)
+                }}
+            >
+                <Icon data={isInAddCategoryMode ? close_circle_outlined : add_circle_outlined} size={16} />
+                <Typography group="navigation" variant="menu_title" as="span">
+                    {isInAddCategoryMode ? 'Close category view' : 'Add project categories'}
+                </Typography>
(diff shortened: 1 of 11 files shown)


frontend/src/views/Admin/BarrierMenu.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { RefObject } from 'react'
+import { Icon, Menu, Typography } from '@equinor/eds-core-react'
+import { add_box, close_circle_outlined } from '@equinor/eds-icons'
```
maiaht, line 3: It does not look bad to use these icons, but why not use a circle for the add icon as well?

frontend/src/views/Admin/QuestionTemplateMenu.tsx
```diff
@@ -1,14 +1,16 @@
...
 interface Props {
     isOpen: boolean
     anchorRef: RefObject<HTMLButtonElement>
     closeMenu: () => void
+    setIsInAddCategoryMode: (inEditmode: boolean) => void
+    isInAddCategoryMode: boolean
 }
 
-const QuestionTemplateMenu = ({ isOpen, anchorRef, closeMenu }: Props) => {
+const QuestionTemplateMenu = ({ isOpen, anchorRef, closeMenu, setIsInAddCategoryMode, isInAddCategoryMode }: Props) => {
```
maiaht, line 13: These changes seem to be unnecessary? Neither the icons nor the two new props are used for anything in this component

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -1,50 +1,122 @@
-import React, { useRef, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
```
maiaht, line 1: `useEffect` is not used in this file

maiaht (approved):

PR #677, opened 2021-09-24
Move page objects into page objects folder

petterwildhagen:
Move page objects which are code interacting with specific elements or elements on specific pages into page_objects folder


frontend/cypress/support/common.ts
```diff
@@ -95,6 +31,7 @@ Cypress.Commands.add('testCacheAndDB', (testCache: Function, testDB: Function =
     cy.reloadBmt()
     testDB()
 })
+export {}
```
achaikou, line 34: What is this?

I can't find explanation in any language references :sweat_smile:

achaikou (approved):
I like that! :heart: 
Just probably squash the weird commits before merging.

frontend/cypress/support/common.ts
petterwildhagen: It says the module is exported which is necessary to make declare global work

petterwildhagen: It indicates that the module is exported, which is necessary to make the declare global work

PR #683, opened 2021-09-24
Create new project category

denektenina:
**Issue**: Make it possible to create a new project category through the admin panel.

**The solution flow:**

1) Click "Add project category" button at the top right

image

2) In the side panel that opens, type a name for the project category. It is not possible to click the save-button if the field is empty, and if the field is erased after typing something, the field will get an error message in addition to the disabled save-button. It is also possible to select a project category in the field below, but let's look at that later. After typing a name, click save.

image

3) After clicking save, the side panel will close, and the new project category will be selected in the project category dropdown. Since the category doesn't have any questions yet, a message will be visible, describing how to get questions into the category. All the same barriers are visible.

image

4) As stated in the text, it is possible to add new questions, and a new feature from this PR is that they will automatically be added to the new project category, since this is selected in the dropdown:

image

image

(notice the "Butterfly" tag on the question template, and the fact that it is the only question listed as Butterfly is selected)

5) It is also possible to add a new project category from a "copy" of another project category. This means that all question templates that is tagged with the other project category, will now be tagged with the new project category. To achieve this we select the project category from the dropdown in the create-sideview:

image

6) As we can see, the new project category has the same question as the one we made earlier (notice the tags):

image

To sum up, this PR gives:
* Possibility to create a new project category in the UI
* Possibility to make a new project category that has all the question templates that a selected other project category has
* New question templates are now assigned to the selected project category

### DIFF
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -282,7 +282,7 @@ namespace api.GQL
 
         public ProjectCategory CopyProjectCategory(string newName, string projectCategoryId)
         {
-            var other = _projectCategoryService.Get(projectCategoryId);
+            var other = _projectCategoryService.GetAll().Include(x => x.QuestionTemplates).Single(x => x.Id ==projectCategoryId);
             return _projectCategoryService.CopyFrom(newName, other);
         }
 
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -286,9 +286,16 @@ namespace api.GQL
             return _projectCategoryService.CopyFrom(newName, other);
         }
 
-        public QuestionTemplate CreateQuestionTemplate(Barrier barrier, Organization organization, string text, string supportNotes)
+        public QuestionTemplate CreateQuestionTemplate(Barrier barrier, Organization organization, string text, string supportNotes, string[] projectCategoryIds)
         {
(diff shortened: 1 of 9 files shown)


backend/api/GQL/Mutation.cs
```diff
@@ -282,13 +282,20 @@ public ProjectCategory CreateProjectCategory(string name)
...
-        public QuestionTemplate CreateQuestionTemplate(Barrier barrier, Organization organization, string text, string supportNotes)
+        public QuestionTemplate CreateQuestionTemplate(Barrier barrier, Organization organization, string text, string supportNotes, string[] projectCategoryIds)
         {
-            return _questionTemplateService.Create(barrier, organization, text, supportNotes);
+            var qt = _questionTemplateService.Create(barrier, organization, text, supportNotes);
+
+            foreach (var projectCategoryId in projectCategoryIds)
+            {
+                _questionTemplateService.AddToProjectCategory(qt.Id, projectCategoryId);
+            }
```
maiaht, line 296: Maybe this code should be in the `Create` function in `QuestionTemplateService.cs`?

denektenina: This was the way Erlend suggested :) Maybe talk to him?

ErlendHaa: Services should need to know about each other, so keeping it here makes more sense

maiaht (approved):

PR #689, opened 2021-09-27
Create new question template

petterwildhagen:


frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -50,10 +59,54 @@ describe('Admin page', () => {
...
+    it('Add new question, fill in title, support notes, select organization, verify question was added', () => {
+        adminPage.createNewQuestion().click()
+        const title = 'title ' + faker.random.word()
+        adminPage.newQuestionTitle().type(title)
+        const supportNotes = 'supportNotes' + faker.random.words(3)
+        adminPage.setSupportNotes(supportNotes)
+        const organization = faker.random.arrayElement(Object.keys(Organization))
+        adminPage.openSelectOrganizationNewQuestion().click()
+        adminPage.selectOrganization(organization).click()
+        adminPage.saveEditButton().click()
```
maiaht, line 71: Maybe this button should be called something else since it is used for both editing and creation?

petterwildhagen: saveQuestion ?

petterwildhagen: Fixed

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -2,36 +2,45 @@ import { users } from '../support/mock/external/users'
...
             })
         })
     })
 
     it('Cancel Edit question - changes to title, support notes and organization are not saved and original values visible', () => {
-        adminPage.allQuestions().then(questions => {
-            var questionsCount = Cypress.$(questions).length
-            const questionNo = faker.datatype.number({ min: 1, max: questionsCount })
+        adminPage.allQuestionNo().then(questions => {
+            var questionsCount = Cypress.$(questions).length - 1
```
hnformentin, line 40: Should this be a const instead of a var? 
As it is only used once, why not to add it directly there?

frontend/cypress/page_objects/admin_page.ts
```diff
@@ -66,4 +70,20 @@ export class AdminPage {
...
+    createNewQuestion = () => {
+        return cy.getByDataTestid('create-new-question')
+    }
+
+    newQuestionTitle = () => {
+        return cy.getByDataTestid('new-question-title')
+    }
+
+    openSelectOrganizationNewQuestion = () => {
+        return cy.getByDataTestid('select-organization').children().eq(0).children().eq(0).children().eq(0).children().eq(0).next()
```
hnformentin, line 83: This is quite not readable. Can you think about an alternative solution?

hnformentin:
Review commit messages: 
We agreed upon writing commit messages in imperative mode during the last "Best practices" meeting. 
This article is also a good read on that : https://chris.beams.io/posts/git-commit/

hnformentin:
Did you run the tests locally in the individual commits to make sure each of them are independent from the others?

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Fixed.

frontend/cypress/page_objects/admin_page.ts
petterwildhagen: Fixed. Added new custom command to extract first child of element

petterwildhagen:
> Did you run the tests locally in the individual commits to make sure each of them are independent from the others?

Yes

frontend/cypress/page_objects/admin_page.ts
```diff
@@ -18,8 +18,8 @@ export class AdminPage {
         return cy.get('button').contains('Admin')
     }
 
-    allQuestions = () => {
-        return cy.get('[id^=question-]')
+    allQuestionNo = () => {
+        return cy.get('[data-testid^=no-question-]')
```
hnformentin, line 22: What does `^` mean here?

frontend/cypress/page_objects/admin_page.ts
petterwildhagen: Fixed by using Dropdown component

frontend/cypress/page_objects/admin_page.ts
petterwildhagen: It means the value of the attribute has to start with what is specified after '='

hnformentin:
> > Did you run the tests locally in the individual commits to make sure each of them are independent from the others?
> 
> Yes

I had now some time to check it, but here there are tests failing in the first commit ("Added test for adding question").
For example, if you do `git checkout 70b2b2d` to your first commit and run `npm run cyrun`, there will be a couple of tests failing.
It seems that it uses the data-testid `create-new-question` but this one is only inserted in the third commit ("Renamed saveQuestion to saveQuestionButton").

Not sure if this was clear, but if you need help to review and test the commits you are creating in this PR, let me know.

frontend/src/views/Admin/AdminView.tsx
```diff
@@ -171,7 +171,12 @@ const AdminView = ({}: Props) => {
                             />
                         </Box>
                         <Box mt={2.5}>
-                            <Button variant="ghost" color="primary" onClick={() => setIsAddingQuestion(true)}>
+                            <Button
+                                variant="ghost"
+                                color="primary"
+                                onClick={() => setIsAddingQuestion(true)}
+                                data-testid="create-new-question"
```
hnformentin, line 178: What about "create-new-question-button"? I saw in other parts of the code and think it is more readable when we know which kind of element we are searching for.

hnformentin: Same suggestion for the remaining data-testid created in this PR...

petterwildhagen: This is the first step to get rid of the page_object classes

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -50,10 +58,61 @@ describe('Admin page', () => {
...
+            adminPage.cancelEditButton().click()
+            cy.testCacheAndDB(() => {
+                goToQuestionnaire()
+                adminPage.allQuestionNo().then(questions => {
+                    var questionsCountPostCancel = Cypress.$(questions).length - 1
+                    if (questionsCount !== questionsCountPostCancel) {
+                        throw new Error('Cancel create new question failed, number of questions before and after cancel edit differ')
+                    } else {
+                        cy.log('Cancel create question verfication OK ' + questionsCountPostCancel + ' === ' + questionsCountPostCancel)
+                    }
```
achaikou, line 105: Maybe use default assertions instead?
`expect(questionsCount, "number of questions before and after cancel edit differ").to.equal(questionsCountPostCancel)`

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -50,10 +58,61 @@ describe('Admin page', () => {
...
+
+    it('Cancel create new question, verify new question is not added', () => {
+        adminPage.allQuestionNo().then(questions => {
+            var questionsCount = Cypress.$(questions).length - 1
+            adminPage.createNewQuestion().click()
+            adminPage.newQuestionTitle().type(faker.random.word())
+            adminPage.setSupportNotes(faker.random.words(3))
+            const dropdownSelect = new DropdownSelect()
+            dropdownSelect.select(
+                cy.getByDataTestid('select-organization-dropdown').contains('Organization'),
```
achaikou, line 93: I still vote for putting at least
`cy.getByDataTestid('select-organization-dropdown').contains('Organization')`
from here and other place into page_objects/admin_page as organizationInput or something

And I personally like page_object classes! :smile: So maybe we need to discuss it more tomorrow :)

achaikou (commented):
I like it in general :+1:

frontend/src/views/Admin/AdminView.tsx
petterwildhagen: Fixed.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -2,36 +2,44 @@ import { users } from '../support/mock/external/users'
 import * as faker from 'faker'
 import { Organization } from '../../src/api/models'
 import { AdminPage } from '../page_objects/admin_page'
+import { DropdownSelect } from '../page_objects/common'
+import { drop } from 'cypress/types/lodash'
```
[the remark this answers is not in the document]
petterwildhagen: deleted

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: git push

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Is organizationInput more descriptive and easier to understand than cy.getByDataTestid('select-organization-dropdown').contains('Organization') ?
Is there a benefit to create an extra abstraction layer in this particular case?

achaikou: I believe there still is.
- avoiding duplication: you already use this string twice. If we decide to change something in the code, it would be faster and easier to do it in one place then to hunt down every one of them, which in theory can happen if bmt grows and grows.
- consistency: it's easier to find elements if they all are located in one location
- readability: I don't really know what `cy.getByDataTestid('select-organization-dropdown').contains('Organization')` means. I am searching for some element called Organization in dropdown that is already supposed to be organization? What? My best guess would be that Organization dropdown contains is a select option called Organization. But that's not true. Something like `adminPage.organizationInput` is more neutral and self-explanatory

petterwildhagen: One solution (which removes duplication and makes page_objects unnecessary) is to create constants in the test itself, i.e.
const organizationInput = 'select-organization-dropdown'

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -102,7 +102,9 @@ const StaticQuestionItem = ({ question, setIsInEditmode, projectCategories, isIn
             <Box display="flex" flexDirection="row">
                 <Box display="flex" flexGrow={1} mb={3} mr={5}>
                     <Box ml={2} mr={1}>
-                        <Typography variant="h4">{question.order}.</Typography>
+                        <Typography variant="h4" data-testid={'question-number-' + question.order}>
```
hnformentin, line 105: Should end as `box`?

frontend/src/views/Admin/CreateQuestionItem.tsx
```diff
@@ -80,7 +82,7 @@ const CreateQuestionItem = ({ setIsAddingQuestion, barrier, questionTitleRef, se
                     </Box>
                 </Box>
                 <Box display="flex" flexDirection={'column'}>
-                    <Box flexGrow={1}>
+                    <Box flexGrow={1} data-testid="select-organization-dropdown">
```
hnformentin, line 85: Should end with `box`?

frontend/src/views/Admin/StaticQuestionItem.tsx
petterwildhagen: I choose not to apply anything because this data is just displayed (not being a markdown editor, dropdown list or something distinctly different). So when only displayed, the naming should be only what is displayed, which in this case is the question number

frontend/src/views/Admin/CreateQuestionItem.tsx
petterwildhagen: The user sees a box, so yes. Fixed.

petterwildhagen:
> I had now some time to check it, but here there are tests failing in the first commit ("Added test for adding question"). For example, if you do `git checkout 70b2b2d` to your first commit and run `npm run cyrun`, there will be a couple of tests failing. It seems that it uses the data-testid `create-new-question` but this one is only inserted in the third commit ("Renamed saveQuestion to saveQuestionButton").
> 
> Not sure if this was clear, but if you need help to review and test the commits you are creating in this PR, let me know.

Thank you for the feedback. I have must have forgotten to run the tests between the commits. I will try to make sure it does not happen again.

achaikou (approved):
Considering that I kinda approved it yesterday let say that things from todays Developer Practices discussion do not apply yet :smile:

PR #691, opened 2021-09-28
Add mutation to delete (void) question template

maiaht:


backend/api/Services/QuestionTemplateService.cs
```diff
@@ -94,6 +94,15 @@ Status status
             return newQuestionTemplate;
         }
 
+        public QuestionTemplate Delete(QuestionTemplate questionTemplate)
+        {
+            ReorderQuestionTemplate(questionTemplate);
```
ErlendHaa, line 99: Maybe add a comment here to explain the logic behind the reorder. Its not obvious why its here

backend/tests/Services/QuestionTemplateService.cs
```diff
@@ -96,6 +96,25 @@ public void EditQuestionTemplate()
...
+                text:         Randomize.String(),
+                supportNotes: Randomize.String()
+            );
+            var projectCategory  = new ProjectCategoryService(_context).GetAll().First();
+            questionTemplateToDelete = service.AddToProjectCategory(questionTemplateToDelete.Id, projectCategory.Id);
+
+            var deletedQuestionTemplate = service.Delete(questionTemplateToDelete);
+
+            Assert.True(deletedQuestionTemplate.Status == Status.Voided);
+            Assert.Equal(deletedQuestionTemplate.ProjectCategories, questionTemplateToDelete.ProjectCategories);
```
ErlendHaa, line 115: Add a test to verify the ordering

backend/api/Services/QuestionTemplateService.cs
maiaht: Fixed

backend/tests/Services/QuestionTemplateService.cs
maiaht: Fixed

ReedOnly (approved):

PR #692, opened 2021-09-28
Add mutation to delete project category

maiaht:


backend/tests/Services/ProjectCategoryService.cs
```diff
@@ -31,6 +31,25 @@ public void Create()
...
+            int nCategories = service.GetAll().Count();
+            var oldProjectCategory = service.GetAll().First();
+            var newProjectCategory = service.CopyFrom(Randomize.String(), oldProjectCategory);
+
+            var questionTemplate = oldProjectCategory.QuestionTemplates.ToList()[0];
+            Assert.True(questionTemplate.ProjectCategories.Contains(newProjectCategory));
+
+            service.Delete(newProjectCategory);
+
+            questionTemplate = oldProjectCategory.QuestionTemplates.ToList()[0];
```
ErlendHaa, line 47: ICollection does not have any guarantees on ordering, so rely on it. It will work in this case as it is the same collection, but still.

Swap line 42 for:
```
var qtService = new QuestionTemplateService(_context);
var qt  = qtService.GetAll().Include(x => x.ProjectCategories).First(x => x.ProjectCategories.Contain(newProjectCategory)
```

then here do:
```
qt  = qtService.GetQuestionTemplate(qt.Id);
```

maiaht: Fixed

backend/tests/Services/ProjectCategoryService.cs
```diff
@@ -31,6 +33,30 @@ public void Create()
...
+            var oldProjectCategory = service.GetAll().First();
+            var newProjectCategory = service.CopyFrom(Randomize.String(), oldProjectCategory);
+
+            var qtService = new QuestionTemplateService(_context);
+            var questionTemplate  = qtService
+                .GetAll()
+                .Include(x => x.ProjectCategories)
+                .First(x => x.ProjectCategories.Contains(newProjectCategory))
+            ;
+            Assert.True(questionTemplate.ProjectCategories.Contains(newProjectCategory));
```
ErlendHaa, line 50: This one is pointless now

maiaht: Fixed

ReedOnly (approved):

PR #696, opened 2021-09-29
Only copy active question templates to a new project category

maiaht:


backend/api/GQL/Mutation.cs
```diff
@@ -282,7 +282,7 @@ public ProjectCategory CreateProjectCategory(string name)
 
         public ProjectCategory CopyProjectCategory(string newName, string projectCategoryId)
         {
-            var other = _projectCategoryService.Get(projectCategoryId);
+            var other = _projectCategoryService.GetAll().Include(x => x.QuestionTemplates).Single(x => x.Id ==projectCategoryId);
```
ErlendHaa, line 285: This doesn't belong here. It related to a different bug, which is already fixed here: https://github.com/equinor/fusion-bmt/pull/683/commits/f8210e897cc6891832f23d16ab30eb7ad8f0551f

maiaht: Fixed

ReedOnly (approved):

PR #698, opened 2021-09-29
Add authorization to admin mutation

### LINKED ISSUE #636: Backend: Restrict mutation in admin page for users with admin privilege
Only let user with Role.Admin perform admin mutations

ReedOnly:
This can only be performed by Admins defined in Azure
AD, controlled by the group FUSION - BMT Administrator

closes #636

### DIFF
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -288,7 +288,7 @@ namespace api.GQL
 
         public ProjectCategory CopyProjectCategory(string newName, string projectCategoryId)
         {
-            var other = _projectCategoryService.GetAll().Include(x => x.QuestionTemplates).Single(x => x.Id ==projectCategoryId);
+            var other = _projectCategoryService.GetAll().Include(x => x.QuestionTemplates).Single(x => x.Id == projectCategoryId);
             return _projectCategoryService.CopyFrom(newName, other);
         }
 
--- a/backend/api/GQL/Mutation.cs
+++ b/backend/api/GQL/Mutation.cs
@@ -278,28 +278,30 @@ namespace api.GQL
          * this role is different from the role used in the
          * application
          */
+        private const string adminRole = "Role.Admin";
 
-        [Authorize(Roles = new [] {"Role.Admin"})]
+
+        [Authorize(Roles = new[] { adminRole })]
         public ProjectCategory CreateProjectCategory(string name)
         {
             return _projectCategoryService.Create(name);
         }
 
-        [Authorize(Roles = new [] {"Role.Admin"})]
+        [Authorize(Roles = new[] { adminRole })]
         public ProjectCategory DeleteProjectCategory(string projectCategoryId)
         {
             var projectCategory = _projectCategoryService.Get(projectCategoryId);
             return _projectCategoryService.Delete(projectCategory);
         }
 
-        [Authorize(Roles = new [] {"Role.Admin"})]
+        [Authorize(Roles = new[] { adminRole })]
         public ProjectCategory CopyProjectCategory(string newName, string projectCategoryId)
         {
             var other = _projectCategoryService.GetAll().Include(x => x.QuestionTemplates).Single(x => x.Id == projectCategoryId);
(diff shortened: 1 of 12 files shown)


backend/api/GQL/Mutation.cs
```diff
@@ -272,26 +273,33 @@ public ClosingRemark CreateClosingRemark(string actionId, string text)
...
          *
-         * There are no role based restictions to these mutations as the
-         * concept "Role" only exists withing an Evaluation.
+         * These mutations will check if the user has the
+         * user role Role.Admin in the jwt token. Note that
+         * this role is different from the role used in the
+         * application
          */
+
+        [Authorize(Roles = new [] {"Role.Admin"})]
```
ErlendHaa: Seams like you 1) ran format on this file 2) made non-formatted changes that where commitet 3) made a new commit that formats you changes. Or something to that effect

backend/api/Services/ProjectCategoryService.cs
```diff
@@ -52,7 +52,7 @@ public ProjectCategory CopyFrom(string newName, ProjectCategory other)
             var newProjectCategory = _Create(newName);
             var activeTemplates = other.QuestionTemplates.Where(qt => qt.Status == Status.Active);
 
-            foreach(var template in activeTemplates)
+            foreach (var template in other.QuestionTemplates)
```
maiaht, line 55: It should be `activeTemplates`, not `other.QuestionTemplates`

maiaht (approved):

backend/api/Startup.cs
```diff
@@ -205,6 +205,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
              * Locally comment this line out to disable authorization. It will
              * allow you to connect to grahpql endpoints and run "schema"
              * command.
+
+             * For Admin mutations the @authorize directive for the given
+             * mutation also has to be commented out.
```
ErlendHaa, line 210: Could you do:
```
var adminRole = new[] {"Role.Admin"} 
```
and use that one in all the mutations? I.e: `[Authorize(adminRole)]`  

Makes it easier to change stuff

backend/api/GQL/Mutation.cs
```diff
@@ -351,6 +362,7 @@ string newNextQuestionTemplateId
             }
         }
 
+        [Authorize(Roles = new[] { "Role.Admin" })]
```
ErlendHaa, line 365: Maybe extend MockAuthService and test that this works as expected?

backend/api/GQL/Mutation.cs
ReedOnly: Yes, I rearranged to formatting to be first now

backend/api/Startup.cs
ReedOnly: The parameters in the attribute needs to be constant, and since new[] is not constant I added
```
private const string adminRole="Role.Admin";
[Authorize(Roles = new [] {adminRole})]
```
instead

backend/api/GQL/Mutation.cs
ReedOnly: We do already, since the mock user penguin already has the role User.Admin which is used by the frontend to decide who will be able to enter the admin menu

frontend/cypress/auth_server/config.json
```diff
@@ -15,7 +15,7 @@
                         "givenName": "Fat",
                         "name": "Fat Cat",
                         "oid": "111",
-                        "roles": ["Reader"],
+                        "roles": ["Role.User"],
```
ReedOnly, line 18: reader role does not exist anymore, I replace all common users with Role.User

backend/api/GQL/Mutation.cs
ErlendHaa: You still format in both commit

backend/api/GQL/Mutation.cs
ReedOnly: I'll create an issue to test this with gql mutation from cypress

PR #706, opened 2021-09-29
Delete question template

denektenina:


frontend/src/views/Admin/Components/ErrorMessage.tsx
```diff
@@ -0,0 +1,22 @@
+import { tokens } from '@equinor/eds-tokens'
+import { apiErrorMessage } from '../../../api/error'
+import { Typography } from '@equinor/eds-core-react'
+
+interface Props {
+    text: string
+}
+
+const ErrorMessage = ({ text }: Props) => {
```
maiaht, line 9: There is already a file called `ErrorSavingQuestion.tsx` that is very similar to this file. However, this is easier to re-use, so maybe delete `ErrorSavingQuestion.tsx` and replace the usage of that component with this component?

denektenina: Good idea, I'll look into it!

denektenina: Fixed

maiaht (approved):

Issue #638, opened 2021-09-15
Remove documentation of EvaluationSeed with corresponding behaviour


hnformentin:
The current documentation in `evaluation_seed.ts` is not updated with the code behaviour.
At least one sample of mismatch was observed:

```
 * The class EvaluationSeed is a specification for the testdata.
 * EvaluationSeed provides methods for adding testdata.
 * Creation of randomized testdata is left to methods defined outside this class.
 * NOTE: The method AddParticipant enforces that the first participant
 * has the role facilitator. See documentation of plant() method for more.
 * The constructor of EvaluationSeed calls addParticipant to create participants from the users.
```

PR #707, opened 2021-09-29
Remove doc about how seed and plant works, let what it is

hnformentin:
Solves #638 

For me some parts of the documentation are useful, and I took away the parts the specify how the method works (and can be read in the code itself).


petterwildhagen (approved):
Looks good.

PR #709, opened 2021-10-01
Reorder question templates in frontend

maiaht:


frontend/src/views/Admin/AdminView.tsx
```diff
@@ -141,8 +149,14 @@ const AdminView = () => {
...
                                         projectCategories={projectCategories}
                                         isInAddCategoryMode={isInAddCategoryMode}
+                                        isInReorderMode={isInReorderMode}
                                         questionTitleRef={questionTitleRef}
                                         refetchQuestionTemplates={refetchQuestionTemplates}
+                                        lowestBarrierOrder={lowestBarrierOrder}
+                                        highestBarrierOrder={highestBarrierOrder}
+                                        prevQuestion={getPrevQuestion(projectCategoryQuestions, q)}
+                                        nextQuestion={getNextQuestion(projectCategoryQuestions, q)}
+                                        nextNextQuestion={getNextNextQuestion(projectCategoryQuestions, q)}
```
denektenina, line 159: Do you have to implement these variables here and send them? I see that you are using some helpers, and also are calculating it, but since you are not using the variables here you can maybe move them to where they are used? Ideally we should send as few props as possible down the tree. I am not referring to isInReorderMode, since you are sending this two places and thus have to create it here.

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -51,6 +68,30 @@ const StaticQuestionItem = ({ question, setIsInEditmode, projectCategories, isIn
         error: deletingQuestionTemplateError,
     } = useDeleteQuestionTemplateMutation()
 
+    const {
+        reorderQuestionTemplate,
+        loading: reorderingQuestionTemplate,
+        error: reorderingQuestionTemplateError,
```
denektenina, line 74: Do you not have error handling since this variable is never used?

denektenina (commented):
Looks good! I have some comments, and also there are some codacy-comments about using '==' instead of '==='

frontend/src/views/Admin/AdminView.tsx
maiaht: These values are calculated here since `projectCategoriesQuestions` and `sortedBarrierQuestions` are available in this file. Would it be better to send those question lists down the tree, and calculate these values in `StaticQuestionItem`?

frontend/src/views/Admin/StaticQuestionItem.tsx
maiaht: Fixed

frontend/src/views/Admin/AdminView.tsx
denektenina: Yes, if you send the "raw" values down, it will be less props :)

maiaht: Fixed

frontend/src/views/Admin/AdminQuestionItem.tsx
```diff
@@ -46,7 +58,11 @@ const AdminQuestionItem = ({ question, projectCategories, isInAddCategoryMode, q
                     setIsInEditmode={setIsInEditmode}
                     projectCategories={projectCategories}
                     isInAddCategoryMode={isInAddCategoryMode}
+                    isInReorderMode={isInReorderMode}
                     questionTitleRef={questionTitleRef}
+                    refetchQuestionTemplates={refetchQuestionTemplates}
+                    sortedBarrierQuestions={sortedBarrierQuestions}
+                    projectCategoryQuestions={projectCategoryQuestions}
```
denektenina, line 65: I don't think we need both of these either. They are the same questions, just organized/sorted differently, right? Maybe the sorting can be a helper and then you can just send the `projectCategoryQuestions` and sort them again in the other component?

frontend/src/views/Admin/StaticQuestionItem.tsx
```diff
@@ -226,6 +278,13 @@ const StaticQuestionItem = ({ question, setIsInEditmode, projectCategories, isIn
                     </Box>
                 </Box>
             )}
+            {reorderingQuestionTemplateError && (
+                <Box mt={2} ml={4}>
+                    <Box>
+                        <TextArea value={apiErrorMessage('Not able to reorder question templates')} onChange={() => {}} />
+                    </Box>
+                </Box>
+            )}
```
denektenina, line 287: You should maybe use the ErrorMessage-component here?

denektenina (commented):
A couple of more comments

frontend/src/views/Admin/AdminQuestionItem.tsx
maiaht: In order to create the `sortedBarrierQuestions` we need to know the `selectedBarrier`. If only `projectCategoryQuestions` is sent to `StaticQuestionItem`, then we also need to send `selectedBarrier`. So it is not really the sorting that is the problem, it is the filtering.

frontend/src/views/Admin/StaticQuestionItem.tsx
maiaht: Fixed!

frontend/src/views/Admin/AdminQuestionItem.tsx
denektenina: I understand. Resolving.

denektenina (approved):
LGTM

PR #716, opened 2021-10-04
Now possible to delete project category

denektenina:
Looks like this:

image

image

Includes:
* Error handling (the Error box can now be closed if the component that uses it provides an onClose-function)
* A spinner in the confirmation box while saving if the component that uses it provides a loading-variable

### DIFF
--- a/frontend/src/views/Admin/CategoryHeader.tsx
+++ b/frontend/src/views/Admin/CategoryHeader.tsx
@@ -1,10 +1,14 @@
 import React, { useState } from 'react'
 import { Box } from '@material-ui/core'
+import { ApolloError, gql, useMutation } from '@apollo/client'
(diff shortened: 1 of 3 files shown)


frontend/src/views/Admin/CategoryHeader.tsx
```diff
@@ -49,24 +84,57 @@ const CategoryHeader = ({
         setSelectedProjectCategory(projectCategory)
     }
 
+    const onDeleteProjectCategory = () => {
+        deleteProjectCategory(selectedProjectCategory)
+    }
```
maiaht, line 89: Does this have to be a function? Or can `deleteProjectCategory(selectedProjectCategory)` be used directly on line 135?

denektenina: Doesn't really matter, but changed it

maiaht (approved):

PR #719, opened 2021-10-04
Add test for delete question template

petterwildhagen:


frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +105,25 @@ describe('Admin page', () => {
...
+            const questionsCount = Cypress.$(questions).length - 1
+            const questionNo = parseInt(
+                questions.toArray()[faker.datatype.number({ min: 0, max: questionsCount })].innerText.replace('.', '')
+            )
+            activeQuestionTemplates().then(activeTemplatesPreDelete => {
+                adminPage.deleteMoveQuestion(questionNo).click()
+
+                adminPage.deleteQuestionTemplate().click({ force: true })
+                adminPage.deleteQuestionTemplateYes().click({ force: true })
+                cy.testCacheAndDB(() => goToQuestionnaire())
```
achaikou, line 119: With the current checks this line makes little sense as all your queries are performed to graphql - you do not interact with the page. You also do nothing here except for navigating to Questionnaire.

You can however check that question text is no longer visible on the page - then reload and navigation would make sense again.
Note, however, that your current goToQuestionnaire reloads the page on its own, so you need to use it only for db-testing, as otherwise you'll loose cache.

Other interesting thing  is numbering. Now if we remove question 2, remaining questions will have order of 1 and 3, but on page reload they change to 1 and 2. That's probably not important, but I am not sure developers made this choice consciously.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +105,25 @@ describe('Admin page', () => {
         })
     })
 
+    it('Delete question template, verify question template was deleted', () => {
+        adminPage.allQuestionNo().then(questions => {
+            const questionsCount = Cypress.$(questions).length - 1
+            const questionNo = parseInt(
+                questions.toArray()[faker.datatype.number({ min: 0, max: questionsCount })].innerText.replace('.', '')
```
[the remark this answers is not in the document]
petterwildhagen: Extracted to specific function . Fixed.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +105,25 @@ describe('Admin page', () => {
...
+            )
+            activeQuestionTemplates().then(activeTemplatesPreDelete => {
+                adminPage.deleteMoveQuestion(questionNo).click()
+
+                adminPage.deleteQuestionTemplate().click({ force: true })
+                adminPage.deleteQuestionTemplateYes().click({ force: true })
```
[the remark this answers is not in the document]
petterwildhagen: deleted button alltogether

frontend/cypress/page_objects/admin_page.ts
```diff
@@ -38,6 +38,18 @@ export class AdminPage {
...
+    }
+
+    deleteQuestionTemplate = () => {
+        return cy.get('span').contains('Delete')
+    }
+
+    deleteQuestionTemplateYes = () => {
+        return cy.getByDataTestid('confirmation_dialog').get('span').contains('Yes')
+    }
+
```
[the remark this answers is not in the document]
petterwildhagen: The GUI changed. Changed page objects & appended 'button' to buttons

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +105,25 @@ describe('Admin page', () => {
...
+    it('Delete question template, verify question template was deleted', () => {
+        adminPage.allQuestionNo().then(questions => {
+            const questionsCount = Cypress.$(questions).length - 1
+            const questionNo = parseInt(
+                questions.toArray()[faker.datatype.number({ min: 0, max: questionsCount })].innerText.replace('.', '')
+            )
+            activeQuestionTemplates().then(activeTemplatesPreDelete => {
+                adminPage.deleteMoveQuestion(questionNo).click()
+
+                adminPage.deleteQuestionTemplate().click({ force: true })
```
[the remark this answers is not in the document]
petterwildhagen: Removed force:true. Fixed

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: I interact with the page in that I delete a question template. Then I count the number of questions in the backend before and after delete and compare the counts.
Regarding the number, not sure what the requirement is here and what the numbers mean. We can discuss that tomorrow in the meetign with them.

frontend/cypress/support/mock/external/external.ts
```diff
@@ -86,7 +86,11 @@ Cypress.on('uncaught:exception', (err, runnable, promise) => {
             const message = messageMatch[1]
 
             /* It looks like majority of our stability issues fall under this message: */
-            if (message === 'Failed to fetch' || message === 'NetworkError when attempting to fetch resource.') {
+            if (
+                message === 'Failed to fetch' ||
+                message === 'NetworkError when attempting to fetch resource.' ||
+                message.includes('Index must be within the bounds of the List.')
```
achaikou, line 92: What is that, why is it here? If that's  an error from our code, it must be fixed.
If that's error from fusion, it should be in its own "if" with own explanations. This section is about "failed to fetch" only.

I don't get this error in tests though, so lets figure out first what causes it.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +100,20 @@ describe('Admin page', () => {
         })
     })
 
+    it('Delete question template, verify question template was deleted', () => {
+        adminPage.allQuestionNo().then(questions => {
+            const questionNo = getRandomQuestionNo(questions, Cypress.$(questions).length - 1)
+            activeQuestionTemplates().then(activeTemplatesPreDelete => {
+                adminPage.deleteQuestionButton(questionNo).click()
+                adminPage.yesButton().click()
+                goToQuestionnaire()
```
achaikou, line 109: > I interact with the page in that I delete a question template. Then I count the number of questions in the backend before and after delete and compare the counts.

If I remove this line, test doesn't fail. Which means it is not used. You interact with backend only, you never check for any changes on the page itself.

If that's intended - simply remove this line.
If not - add some checks - like search for a question text on the page to make sure it is removed from cache also.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -104,6 +100,20 @@ describe('Admin page', () => {
         })
     })
 
+    it('Delete question template, verify question template was deleted', () => {
+        adminPage.allQuestionNo().then(questions => {
+            const questionNo = getRandomQuestionNo(questions, Cypress.$(questions).length - 1)
+            activeQuestionTemplates().then(activeTemplatesPreDelete => {
+                adminPage.deleteQuestionButton(questionNo).click()
+                adminPage.yesButton().click()
```
achaikou, line 108: > deleted button alltogether

Not exactly.

OK, one more time and I drop the topic as it's not really important. :cat: 

All the confirmation stuff in the source code is handled by `frontend/src/components/ConfirmationDialog.tsx`
That's why I believe that it would be better to kinda use one page object for this component in tests also
```
import { ConfirmationDialog } from '../page_objects/common'
new ConfirmationDialog().yesButton()
```

But that's a small locator, so lets say that I can survive with it being duplicated if you are not fond of this "ConfirmationDialog" thing...

frontend/cypress/support/mock/external/external.ts
petterwildhagen: Good catch! This is introduced to handle getting elements outside of valid boundaries. I will remove it.

PR #721, opened 2021-10-05
Add test regular user cannot see admin tab and cannot delete question template using mutations

petterwildhagen:


frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -3,13 +3,15 @@ import * as faker from 'faker'
...
+import { DELETE_QUESTION_TEMPLATE } from '../support/testsetup/gql'
+import { activeQuestionTemplates } from '../support/testsetup/evaluation_seed'
 
 const adminPage = new AdminPage()
 const selectOrganizationDropdown = 'select-organization-dropdown-box'
 describe('Admin page', () => {
     const goToQuestionnaire = () => {
-        const user = users[users.length - 1]
-        cy.visitProject(user)
+        const adminUser = users[users.length - 1]
```
achaikou, line 13: Can we do it better, like by filtering users who have Admin privilege? Now we rely too much on correct user placement.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -110,6 +112,28 @@ describe('Admin page', () => {
...
     })
+
+    it('Non admin user does not see Admin tab', () => {
+        const nonAdminUser = users[users.length - 2]
+        cy.visitProject(nonAdminUser)
+        adminPage.adminButton().should('not.exist')
+    })
+
+    it('Non admin user cannot do GQL mutations to delete question', () => {
+        const nonAdminUser = users[users.length - 2]
```
achaikou, line 123: Same in both places here. It would be nice to just get any user who does not have admin privilege and not to rely on placement.

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -110,6 +112,28 @@ describe('Admin page', () => {
...
+        cy.visitProject(nonAdminUser)
+        activeQuestionTemplates().then(templates => {
+            const idOfFirstTemplate = templates[0].id
+            cy.gql(DELETE_QUESTION_TEMPLATE, { variables: { questionTemplateId: idOfFirstTemplate } }).then(res => {
+                const errorCode = res.body.errors[0].extensions.code
+                const errorMessage = res.body.errors[0].message
```
achaikou, line 129: Maybe you should use here safe navigation operator? `errors?.[0]` Otherwise if I run this code with admin everything fails already here and doesn't get to your nice code/message comparison

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -110,6 +112,28 @@ describe('Admin page', () => {
...
+                const expectedErrorCode = 'AUTH_NOT_AUTHORIZED'
+                const expectedErrorMessage = 'The current user is not authorized to access this resource.'
+                expect(errorCode, `error code is not ${expectedErrorCode}`).to.equal(expectedErrorCode)
+                expect(errorMessage, `error message is not ${expectedErrorMessage}`).to.equal(expectedErrorMessage)
```
achaikou, line 133: Not sure if we need to test message too. I think that message comes from ChilliCream thing https://github.com/ChilliCream/hotchocolate/search?q=The+current+user+is+not+authorized+to+access+this+resource.

I also see the point in having it here, but maybe error code is enough if we agree that we want to test as little external code as possible.

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Fixed. Added function to get user with Admin role

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: Fixed. Added function to get user with no admin role

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: This code should not be run with admin, only with user with no admin role. If run with admin role then it should fail already here. This code only makes sense if run with no admin role. So I think it is better to leave as is.

frontend/cypress/integration/admin_page_spec.ts
petterwildhagen: I discussed this with @ErlendHaa  and we agreed to include the text from the error message. But I can also delete it. We can discuss it tomorrow on standup.

frontend/cypress/integration/admin_page_spec.ts
achaikou: OK, we can leave it, I am not going to argue much, but what's the point in nice "expects" then?

Nice assertions are supposed to fail if something breaks and our non-admin users suddenly get access to this mutation. But in this case they will fail here (which is at the moment easily testable by running this test by admin) with just syntax message and won't give us any good information about what went wrong.

So that is a little weird to me.
If you want to we can discuss it a bit on standup tomorrow along with another point.

petterwildhagen: I have added asserts with expect to check for presence of structure. Is that better?

achaikou (approved):

PR #723, opened 2021-10-05
Add test to delete prject category

petterwildhagen:


frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -151,6 +153,34 @@ describe('Admin page', () => {
...
+            adminPage.allQuestionNo().then(questions => {
+                expect(questions.length, ' number of questions should be ').to.equal(1)
+            })
+            adminPage.questionNoByTitle(questionTitle).then(qNo => {
+                const questionNo = parseInt(Cypress.$(qNo).text())
+                verifyQuestion(questionNo, questionTitle, organization, supportNotes)
+                adminPage.deleteProjectCategory().click()
+                new ConfirmationDialog().yesButton().click()
+                adminPage.selectProjectCategoryDropdown().click()
+                dropdownSelect.assertSelectValues(['CircleField', 'SquareField', 'All project categories'])
```
achaikou, line 179: Hm... This expects full list of expected values. Thus if some other future test adds the category without removing it, this will fail.
Would it make sense to get all existing project categories before the start of the test? Or do absolutely anything else that guards against it?

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -151,6 +153,34 @@ describe('Admin page', () => {
...
+                cy.contains(newCategoryName).should('not.exist')
```
achaikou, line 180: testCacheAndDB? (or again reuse goToQuestionnarie navigation) we will also test it is removed from db

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -151,6 +153,34 @@ describe('Admin page', () => {
...
     })
+
+    it('Select question template by category, delete project category & verify project category is no longer present', () => {
+        const newCategoryName = 'TheNewCategory' + faker.lorem.word()
+        createNewProjectCategory(newCategoryName).then(categoryId => {
+            const questionTitle = faker.lorem.words(2)
+            const organization = faker.random.arrayElement(Object.keys(Organization))
+            const supportNotes = faker.lorem.words(3)
+            // @ts-ignore
+            let org: Organization = Organization[organization]
```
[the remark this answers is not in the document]
petterwildhagen: Fixed

frontend/cypress/page_objects/admin_page.ts
```diff
@@ -26,6 +30,10 @@ export class AdminPage {
         return cy.get('[data-testid^=question-number-]')
     }
 
+    selectProjectCategoryDropdown = () => {
+        return cy.get('[value="All project categories"]')
+    }
+
```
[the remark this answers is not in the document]
petterwildhagen: Either works, but label might be more specific. Fixed

frontend/cypress/support/testsetup/evaluation_seed.ts
```diff
@@ -319,3 +320,15 @@ export function projectCategoryId(categoryName: string): Cypress.Chainable<strin
         return id
     })
 }
+
+export function allProjectCategoryNames(): Cypress.Chainable<Array<string>> {
+    return cy.gql(GET_ALL_PROJECT_CATEGORY_NAMES, { variables: {} }).then(res => {
+        console.log(res)
```
achaikou, line 326: Remove

frontend/cypress/integration/admin_page_spec.ts
```diff
@@ -154,29 +154,31 @@ describe('Admin page', () => {
...
+                })
+                adminPage.questionNoByTitle(questionTitle).then(qNo => {
+                    const questionNo = parseInt(Cypress.$(qNo).text())
+                    verifyQuestion(questionNo, questionTitle, organizationToString(organization), supportNotes)
+                    adminPage.deleteProjectCategory().click()
+                    new ConfirmationDialog().yesButton().click()
+                    adminPage.selectProjectCategoryDropdown().click()
+                    dropdownSelect.assertSelectValues(projectCatArray.concat('All project categories'))
+                    goToQuestionnaire()
+                    cy.contains(newCategoryName).should('not.exist')
```
achaikou, line 180: Here is unfortunate moment - this thing only makes sense when dropdown is open, because looks like this is when DOM is populated... But after you renavigate it is closed, so it will always say it doesn't exist.

So I think you again need 
```
adminPage.selectProjectCategoryDropdown().click()
cy.contains(newCategoryName).should('not.exist')
```

achaikou: Huh, haven't thought about what happens with existing questions when category is deleted. Good that you tested it!

But you misunderstood me a bit.
The code before questionnaire already was good because you checked `ropdownSelect.assertSelectValues(projectCatArray.concat('All project categories'))`
If the category was still present, it would have failed.

But I think you need to add
`adminPage.selectProjectCategoryDropdown().click()`
after you `goToQuestionnaire()`

Because as you say you check that it's not present in any question, but it can be that question is not there, but category is.

PR #731, opened 2021-10-08
Squash order of questions in evaluation

maiaht:


backend/tests/GQL/Mutations/CreateEvaluation.cs
```diff
@@ -38,5 +38,29 @@ public void CreateAddsCorrectQuestions()
...
+                .Where(pc => pc.Name == "SquareField")
+                .First()
+            ;
+
+            var evaluation = _mutation.CreateEvaluation(
+                name: Randomize.String(),
+                projectId: _projectService.GetAll().First().Id,
+                previousEvaluationId: "",
+                projectCategoryId: projectCategory.Id
+            );
```
ErlendHaa, line 56: Re-fetch the evaluation to also test that the changes are properly saved in the db.
`evaluation = _evaluationService.GetEvaluation(evaluation.Id)`
This is something we don't do a lot in other test, but that we should do all the time

PR #739, opened 2021-10-12
Copy question templates

maiaht:


frontend/src/views/Admin/CopyQuestionItem.tsx
```diff
@@ -0,0 +1,190 @@
...
+            variables: { ...data },
+        })
+    }
+
+    return {
+        copyQuestionTemplate,
+        loading,
+        error,
+    }
+}
```
denektenina, line 190: Could be worth a try to reuse the CreateQuestionTemplate-component for this, since there is a lot of reuse?

denektenina (approved):
Looks good to me now, but remember to squash the commits and write a good commit message!

PR #772, opened 2021-10-15
Test that question numbers in evaluations are sequential

petterwildhagen:


frontend/cypress/integration/evaluation_spec.ts
```diff
@@ -60,6 +61,31 @@ describe('Evaluation management', () => {
...
+                            if (q !== 0) {
+                                cy.get(evaluationPage.questionNoSelector).each(q => {
+                                    expect(parseInt(q.text().replace('.', '')), ' question number is in a sequence').to.equal(
+                                        questionCounter
+                                    )
+                                    questionCounter++
+                                })
+                            }
+                            if (element === Object.values(Barrier).slice(-1)[0]) {
+                                expect(totalQuestionCounter, 'All active question question templates were added').eq(
```
maiaht, line 83: "question" is written two times

frontend/cypress/integration/evaluation_spec.ts
```diff
@@ -60,6 +61,31 @@ describe('Evaluation management', () => {
                             expectedTemplates.length
                         )
                     })
+                    cy.visitProgression(Progression.Individual, currentEvaluation.id, user, fusionProject1.id)
+                    cy.contains('Questionaire')
+                    const evaluationPage = new EvaluationPage()
+                    let questionCounter = 1
```
maiaht, line 67: Why does this counter start at 1?

petterwildhagen: Because the numbering of the questions starts at 1. The first time the assertion kicks in for the question number you assert it is 1. The questionCounter is incremented. The second time you assert you assert the question number is 2 etc

PR #773, opened 2021-10-18
In admin_page_spec.ts replace faker.lorem.word(s) with custom function

petterwildhagen:


frontend/cypress/support/testsetup/testdata.ts
```diff
@@ -20,3 +20,15 @@ export function createAction(
...
     return new Action({ assignedTo, createdBy, questionOrder, dueDate, title, priority, description, completed, onHold })
 }
+
+const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
+
+export function generateRandomString(length: number) {
+    let result = ''
+    const charactersLength = characters.length
+    for (let i = 0; i < length; i++) {
+        result += characters.charAt(Math.floor(Math.random() * charactersLength))
```
maiaht, line 30: Since `Math.random()` returns a number in the interval [0,1) and you use `Math.floor()`, I don't think the last letter of characters will ever be a part of the result.

maiaht (approved):

Issue #812, opened 2021-10-25
Devops: Add attempt to cypress build id


ReedOnly:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/run-number?view=azure-devops&tabs=yaml

PR #832, opened 2021-10-28
Adding name to yaml to add rev nr to buildid

eiriksanderfjeld:
Setting the value of name to buildname, buildid and revnr. #812

### DIFF
--- a/pipelines/cypress-E2E.yml
+++ b/pipelines/cypress-E2E.yml
@@ -7,7 +7,7 @@ pr:
 
 pool:
     vmImage: "ubuntu-latest"
-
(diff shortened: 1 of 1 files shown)


lightenup (dismissed):

ReedOnly:
do not merge in this state!

eiriksanderfjeld:
Setting the pipeline name to $(Build.BuildId).$(rev:r), IE 10419.1, and setting the ci-build-id to the same, plus jobattempt, which ends up like this: 10419.1.1.

ReedOnly (approved):

PR #854, opened 2021-11-03
New view for portfolios

denektenina:
image

The hook to get portfolios is changed to optimize loading time, and handle async calls, but the original logic for where to find portfolio is maintained. Needs to be tested because it's difficult to know if the portfolios are correct.

### DIFF
--- a/frontend/cypress/support/mock/external/projects.ts
+++ b/frontend/cypress/support/mock/external/projects.ts
@@ -40,7 +40,26 @@ export const fusionProject6 = new FusionProject({
     name: 'Broken Coconut',
 })
 
-export const fusionProjects = [fusionProject1, fusionProject2, fusionProject3, fusionProject4, fusionProject5, fusionProject6]
+export const fusionProject7 = new FusionProject({
+    id: '1',
+    name: 'Because backend',
+})
+
+export const fusionProject8 = new FusionProject({
+    id: '2',
+    name: 'Backend made me',
+})
+
+export const fusionProjects = [
+    fusionProject1,
+    fusionProject2,
+    fusionProject3,
+    fusionProject4,
+    fusionProject5,
+    fusionProject6,
+    fusionProject7,
+    fusionProject8,
+]
 
 export function findFusionProjectByID(id: string) {
     return fusionProjects.filter(p => p.id == id)[0]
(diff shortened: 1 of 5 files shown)


frontend/src/utils/hooks.ts
```diff
@@ -47,32 +47,96 @@ export const useAllPersonDetailsAsync = (azureUniqueIds: string[]): PersonDetail
...
-    evaluation: Evaluation
-    portfolio: string | undefined
+export type PortfolioByProjectId = {
+    [projectId: string]: string
 }
 
-export const useEvaluationsWithPortfolio = (evaluations: Evaluation[]): EvaluationWithPortfolio[] => {
+export type EvaluationWithPortfolio = {
+    [key: string]: Evaluation[]
+}
```
maiaht, line 56: Should this type be renamed to `EvaluationsWithPortfolio`?

frontend/src/views/Project/Dashboard/ProjectDashboardView.tsx
```diff
@@ -68,23 +74,37 @@ const ProjectDashboardView = ({ project }: Props) => {
...
     } = useProjectEvaluationsQuery(project.id, Status.Active)
 
-    const { loading: loadingHiddenEvaluations, evaluations: hiddenEvaluations, error: errorHiddenEvaluations } = useAllEvaluationsQuery(
-        Status.Voided
-    )
+    const {
+        loading: loadingAllEvaluations,
+        evaluations: allEvaluations,
+        error: errorAllEvaluations,
+    } = useAllEvaluationsQuery(Status.Active)
```
maiaht, line 99: Could these variables be called `activeEvaluations` instead of `allEvaluations`?

frontend/cypress/support/mock/external/projects.ts
```diff
@@ -40,7 +40,26 @@ export const fusionProject6 = new FusionProject({
...
+export const fusionProjects = [
+    fusionProject1,
+    fusionProject2,
+    fusionProject3,
+    fusionProject4,
+    fusionProject5,
+    fusionProject6,
+    fusionProject7,
+    fusionProject8,
+]
```
maiaht, line 62: Why are the changes in this file a part of this PR?

denektenina: Good question, it's because of failing tests. The hook does a couple of calls to `queryContextsAsync`, and when I use the hook in the Dashboard, the tests that loaded the Dashboard failed because the query wasn't intercepted by cypress.

maiaht (approved):

PR #859, opened 2021-11-05
Adding templates to increase readabillity

eiriksanderfjeld:
Use templates for stages, which enables code reusabillity, increased
readabillity and ensures identical pipeline for each environment. #857


ReedOnly (approved):
looks good!

PR #864, opened 2021-11-08
Downgrade Cypress from 8.7 to 8.4

petterwildhagen:
Downgrade Cypress from 8.7 to 8.4 as videos no longer work in 8.7

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -50,7 +50,7 @@
         "@types/styled-components": "^5.1.15",
         "@types/webpack-env": "^1.16.3",
         "babel-jest": "^27.3.1",
-        "cypress": "8.7.0",
+        "cypress": "8.4.0",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -50,7 +50,7 @@
         "@types/styled-components": "^5.1.15",
         "@types/webpack-env": "^1.16.3",
         "babel-jest": "^27.3.1",
-        "cypress": "8.7.0",
+        "cypress": "8.4.0",
```
achaikou, line 53: You also have to update `frontend/package-lock.json ` - it's changed after npm install
And might be good to put in the commit description short info that we are reverting to older version as videos are not playing anymore.

achaikou (approved):

frontend/package.json
petterwildhagen: Fixed

PR #870, opened 2021-11-11
Adding branch to template for radix trigger

eiriksanderfjeld:
To avoid identical id for the deployment step and to ensure radix builds
correct environment we send in branch as parameter. #869

### DIFF
--- a/pipelines/ArtifactBuild.yml
+++ b/pipelines/ArtifactBuild.yml
@@ -145,6 +145,7 @@ stages:
           radixToken: $(RadixToken)
           buildId: $(Build.BuildId)
           envshortname: CI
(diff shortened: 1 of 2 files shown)


pipelines/ArtifactBuild.yml
```diff
@@ -145,6 +145,7 @@ stages:
           radixToken: $(RadixToken)
           buildId: $(Build.BuildId)
           envshortname: CI
+          branch: dev
```
ReedOnly, line 148: should be branch: master

ReedOnly (changes requested):

ReedOnly (approved):

PR #885, opened 2021-11-16
Added url rendring and icon to add url syntax to input field on closi…

denektenina:
…ng comment

There has been a request that it should be possible to add a hyperlink when closing an action. With this PR, a url-button is added close to the closing-reason input field. When clicking this, the reason-field will get the following syntax added: [text](url), so that they can insert the link with the right syntax:

image

Also changed the rendering of comments so that the link will be rendered as a link. This also works for notes now, but the note field doesn't have the link-button as of now, since that was not part of the assignment. It is however possible to use the same syntax to write a link there as well.

For some reason, the `<MarkdownViewer>` added a \n to the comment when fetched from the UI, so the test had to be slightly modified to accommodate this.

### DIFF
--- a/frontend/src/components/Action/EditForm/NotesAndClosingRemarksList.tsx
+++ b/frontend/src/components/Action/EditForm/NotesAndClosingRemarksList.tsx
@@ -3,6 +3,7 @@ import { Box } from '@material-ui/core'
 import { ClosingRemark, Note } from '../../../api/models'
 import { PersonDetails } from '@equinor/fusion'
 import { tokens } from '@equinor/eds-tokens'
+import { MarkdownViewer } from '@equinor/fusion-components'
 
 interface Props {
     notesAndClosingRemarks: (Note | ClosingRemark)[]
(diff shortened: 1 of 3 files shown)


frontend/src/components/Action/EditForm/NotesAndClosingRemarksList.tsx
```diff
@@ -58,8 +59,8 @@ const NotesAndClosingRemarksList = ({ notesAndClosingRemarks, participantsDetail
                                 {hoursPadded}:{minutesPadded} - {dateString}
                             </Box>
                         </Box>
-                        <Box mt={1} data-testid={'note_text'}>
-                            {noteOrRemark.text}
+                        <Box data-testid={'note_text'}>
+                            <MarkdownViewer data-testid={'note_text'} markdown={noteOrRemark.text} />
```
maiaht, line 63: Is it necessary that both the `Box` and the `MarkdownViewer` have `data-testid={'note_text'}`?

denektenina: No, I can remove one

maiaht (approved):

PR #897, opened 2021-11-19
Stabilize failing landing spec test

achaikou:
Ready for review.
It can't be guaranteed that stability issue is solved, but at least it's not made worse - out of 80 runs in 4 batches there were no failures.


petterwildhagen (approved):
Looks good. Lets see how it works.

PR #908, opened 2021-11-23
Sort evaluations by project master

maiaht:
Skjermbilde 2021-11-23 kl  12 30 10 (2)

### DIFF
--- a/frontend/src/utils/hooks.ts
+++ b/frontend/src/utils/hooks.ts
@@ -47,12 +47,17 @@ export const useAllPersonDetailsAsync = (azureUniqueIds: string[]): PersonDetail
     return { personDetailsList, isLoading }
 }
 
-export type PortfolioByProjectId = {
-    [projectId: string]: string
+export type PortfolioAndProjectMasterTitleByProjectId = {
+    [projectId: string]: [string, string]
 }
 
 export type EvaluationsWithPortfolio = {
-    [key: string]: Evaluation[]
+    [key: string]: EvaluationWithProjectMasterTitle[]
+}
+
+export interface EvaluationWithProjectMasterTitle {
+    evaluation: Evaluation,
+    projectMasterTitle: string
 }
 
 export const noPortfolioKey = 'No portfolio'
@@ -73,28 +78,29 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
         return projectIds
     }
 
-    const lookupPortfolio = async (projectId: string) => {
+    const lookupPortfolioAndProjectMaster = async (projectId: string) => {
         const project = await apiClients.context.getContextAsync(projectId)
         const projectMasterId = project.data.value.projectMasterId
 
         if (projectMasterId) {
             const projectMaster = await apiClients.context.queryContextsAsync(projectMasterId, ContextTypes.ProjectMaster)
             const portfolio = projectMaster.data[0].value.portfolioOrganizationalUnit
-            return portfolio
+            const projectMasterTitle = projectMaster.data[0].title
+            return [portfolio, projectMasterTitle]
         }
-        return ''
+        return ['', '']
     }
 
-    const assignPortfoliosToId = async (projectIds: string[]) => {
-        const projectIdsWithPortfolios: PortfolioByProjectId = {}
+    const assignPortfoliosAndProjectMasterTitileToId = async (projectIds: string[]) => {
+        const projectIdsWithPortfolios: PortfolioAndProjectMasterTitleByProjectId = {}
 
         await Promise.all(
             projectIds.map(async projectId => {
                 try {
-                    const portfolio = await lookupPortfolio(projectId)
-                    projectIdsWithPortfolios[projectId] = portfolio
+                    const [portfolio, projectMasterTitle] = await lookupPortfolioAndProjectMaster(projectId)
+                    projectIdsWithPortfolios[projectId] = [portfolio, projectMasterTitle]
                 } catch (err) {
-                    projectIdsWithPortfolios[projectId] = ''
+                    projectIdsWithPortfolios[projectId] = ['', '']
                 }
             })
         )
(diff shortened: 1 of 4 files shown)


frontend/src/utils/hooks.ts
```diff
@@ -73,28 +78,29 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
...
+            const projectMasterTitle = projectMaster.data[0].title
+            return [portfolio, projectMasterTitle]
         }
-        return ''
+        return ['', '']
     }
 
-    const assignPortfoliosToId = async (projectIds: string[]) => {
-        const projectIdsWithPortfolios: PortfolioByProjectId = {}
+    const assignPortfoliosAndProjectMasterTitileToId = async (projectIds: string[]) => {
```
denektenina, line 94: Typo: Titile

frontend/src/utils/hooks.ts
```diff
@@ -111,24 +117,28 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
             // on each projectId instead of each evaluation to save loading time
             const projectIds = collectUniqueProjectIds(evaluations)
 
-            assignPortfoliosToId(projectIds).then(projectIdsWithPortfolios => {
+            assignPortfoliosAndProjectMasterTitileToId(projectIds).then(projectIdsWithPortfolios => {
```
denektenina, line 120: projectIdsWithPortfolios is maybe not the name you are using now?

frontend/src/utils/hooks.ts
```diff
@@ -47,12 +47,17 @@ export const useAllPersonDetailsAsync = (azureUniqueIds: string[]): PersonDetail
     return { personDetailsList, isLoading }
 }
 
-export type PortfolioByProjectId = {
-    [projectId: string]: string
+export type PortfolioAndProjectMasterTitleByProjectId = {
+    [projectId: string]: [string, string]
```
denektenina, line 51: Maybe this should be an object and not array? I.e. `[projectId: string]: {portfolio: string, masterTitle: string}
`
You can still use destructuring further down in the code:

`const { portfolio, masterTitle } = projectIdsWithPortfolios[projectId]`

frontend/src/views/Project/Dashboard/Portfolios.tsx
```diff
@@ -1,24 +1,58 @@
...
+                    if (!uniqueProjectMasterTitlesWithPortfolio.map(a => a.projectMasterTitle).includes(e.projectMasterTitle)) {
+                        uniqueProjectMasterTitlesWithPortfolio.push({
+                            portfolio,
+                            projectMasterTitle: e.projectMasterTitle,
+                        })
+                    }
+                })
+            })
+        setPortfolioAndProjectMasterTitles(uniqueProjectMasterTitlesWithPortfolio)
+    }, [evaluationsWithPortfolio])
```
denektenina, line 38: Should you maybe return this format from the hook instead of reworking it in this component? The hook is not used other places so it can be custom made for the type of rendering you want to do in your UI component.

denektenina (commented):
It's a lot of different object types here and the names are really long which makes the code a bit difficult to read. Also, I feel like we are going through a lot of iterations to create different objects, and at least one of them is unnecessary I think. Like I wrote in a comment, see if you can return the correct format from the hook instead of reworking it in the Portfolios-component :)

I understand if you are getting dizzy working with this, it might help using pen and paper with examples of the objects to gain some clarity! Part of writing good code is to find the simplest way to present the data, later developers will thank you for it! :)

frontend/src/utils/hooks.ts
maiaht: Fixed

frontend/src/utils/hooks.ts
maiaht: Fixed

frontend/src/utils/hooks.ts
maiaht: Fixed

frontend/src/utils/hooks.ts
```diff
@@ -107,36 +124,48 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
...
                     }
                 })
                 setAllEvaluationsWithPortfolio(evaluationsWithPortfolio)
+                setPortfolioAndProjectMasterTitles(uniqueProjectMasterTitlesWithPortfolio)
             })
         }
     }, [evaluations])
 
-    return allEvaluationsWithPortfolio
+    return [allEvaluationsWithPortfolio, portfolioAndProjectMasterTitles]
```
denektenina, line 168: I think this is still too messy and it doesn't reflect the way you are using the data in the frontend. Like I said on chat, you are using filter and map a lot to look up data instead of having it ready where you need it. I would return a data type like this from the hook:

```
type EvaluationsByTitleAndPortfolio = {
    [portfolio: string]: EvaluationsByProjectMaster
}

(type EvaluationsByProjectMaster = {
    [projectMasterTitle: string]: Evaluation[]
})
```

that way you just need to iterate through each portfolio and for each of those, iterate through each ProjectMaster and then list all its evaluations.

image

frontend/src/utils/hooks.ts
```diff
@@ -73,70 +83,83 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
...
+            const projectMasterTitle = projectMaster.data[0].title
+            return [portfolio, projectMasterTitle]
         }
-        return ''
+        return ['', '']
     }
 
-    const assignPortfoliosToId = async (projectIds: string[]) => {
-        const projectIdsWithPortfolios: PortfolioByProjectId = {}
+    const assignPortfoliosAndProjectMasterTitleToId = async (projectIds: string[]) => {
```
denektenina, line 99: You can remove "title" here as well

frontend/src/utils/hooks.ts
```diff
@@ -73,70 +83,83 @@ export const useEvaluationsWithPortfolio = (evaluations: Evaluation[] | undefine
...
+                [noPortfolioKey]: {
+                    [noProjectMasterTitle]: [],
+                },
             }
             // Since many evaluations have the same projectId, and thus the same portfolio, we are only doing lookups
             // on each projectId instead of each evaluation to save loading time
             const projectIds = collectUniqueProjectIds(evaluations)
 
-            assignPortfoliosToId(projectIds).then(projectIdsWithPortfolios => {
+            assignPortfoliosAndProjectMasterTitleToId(projectIds).then(projectIdsWithPortfoliosAndProjectMasterTitles => {
```
denektenina, line 127: And remove "title" from these two

denektenina (approved):
Looks good! I have two name comments, but after that you can merge! I assume you have run it and it looks correct and that tests are passed?

PR #910, opened 2021-11-24
Made some changes to the CreateEvaluationDialog-view, to comply with the way the UI looks and acts in the rest of the application.

denektenina:
- Started to use a new error display for errors in smaller views: ` <ErrorBanner/>`. This will be instead of the grey text fields we have been using until now:

Earlier:
image

Now:
image

- The error display is also moved. Earlier, if creating an evaluation didn't succeed, the sidebar would close, and the old error display would replace the "Create Evaluation"-button on the Dashboard. Now, the error banner will appear in the sidebar, making it possible to try again without losing the typed data. 

Earlier:
image

Now:
image

- Removed the text explaining that it's difficult to remove a created evaluation, since we now have the possibility to hide an evaluation.

image

- Started using the same validation as in other forms in the application. I.e. warning message in the Input-component, instead of a free text next to the save-button. We are also using a hook to update validation status, so that we don't have to duplicate code.

Earlier:
image

Now:
image

- Inserted cancel-button in the view

- Moved both buttons to the right, like we have most places

- Create-button is now disabled when required fields are not filled out

- (Required) behind the label on the required fields

### DIFF
--- a/frontend/src/views/Project/Dashboard/DashboardView.tsx
+++ b/frontend/src/views/Project/Dashboard/DashboardView.tsx
@@ -159,7 +159,9 @@ const DashboardView = ({ project }: Props) => {
             )}
             {portfoliosSelected && (
                 <>
-                    {allActiveEvaluationsWithProjectMasterAndPortfolio && <Portfolios evaluationsWithProjectMasterAndPortfolio={allActiveEvaluationsWithProjectMasterAndPortfolio} />}
+                    {allActiveEvaluationsWithProjectMasterAndPortfolio && (
+                        <Portfolios evaluationsWithProjectMasterAndPortfolio={allActiveEvaluationsWithProjectMasterAndPortfolio} />
+                    )}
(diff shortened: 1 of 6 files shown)


frontend/src/views/Project/Dashboard/DashboardView.tsx
```diff
@@ -162,7 +162,7 @@ const DashboardView = ({ project }: Props) => {
                     {allActiveEvaluationsWithProjectMasterAndPortfolio && <Portfolios evaluationsWithProjectMasterAndPortfolio={allActiveEvaluationsWithProjectMasterAndPortfolio} />}
                     {(loadingActiveEvaluations || !allActiveEvaluationsWithProjectMasterAndPortfolio) && <CenteredCircularProgress />}
                     {errorActiveEvaluations !== undefined && errorMessage}
-                </>
+                   </>
```
maiaht, line 165: Is this change correct?

denektenina: Seems like prettier didn't reformat. Fixed now.

maiaht (approved):
Looks good!

PR #912, opened 2021-11-24
Rename severity

maiaht:


denektenina (approved):
Looks good, and much needed in my opinion! Just make sure LOW is really the same as MAJOR ISSUES, as this is not very intuitive from the code. (Severity.LOW sounds like it's not very severe, but I think that's the problem we have had with this earlier, and will now finally get rid of!)

PR #917, opened 2021-11-29
Refactoring3

denektenina:
Some more refactoring changes:

### CancelAndSaveButton
Added new component `<CancelAndSaveButton>`, that can be used everywhere where those two buttons are used. The Save-button has a spinner that shows whenever the loading-parameter is set to true, so that the user knows what is happening if the saving takes some seconds.

image

The component is in this PR used in the following components:
- CreateProjectCategorySidebar
- CreateQuestionItem
- EditableQuestionItem
- CreateEvaluationDialog

This component replaced the previous component `<CancelOrSaveQuestion>`, that was used only in the Admin panel.

### CircularProgress
Using the EDS `<CircularProgress> `when data is fetched, instead of the "loading..."-text. Changed in the following components:

 - AdminView
 - StaticQuestionItem

### ErrorMessage
Using the Fusion <ErrorMessage>-component when data is missing from a view

image

In this PR this is implemented for the following views:

- AdminView

### ErrorBanner
Using <ErrorBanner> based on Banner from EDS instead of the grey text fields we used for displaying Error earlier. In this PR this is included in the following components:

- CategoryHeader
- CreateProjectCategorySidebar
- CreateQuestionItem
- EditableQuestionItem
- StaticQuestionItem

Edit: Has been tested manually by going through all the points in https://github.com/equinor/fusion-bmt/issues/923 and mocking errors /loading.


frontend/src/views/Project/Admin/Components/EditableQuestionItem.tsx
```diff
@@ -1,51 +1,52 @@
...
+    }, [isQuestionTemplateSaving])
 
-    useEffect(() => {
-        updateValidity(isTextfieldValid(), textValidity, setTextValidity)
-    }, [text, textValidity])
+    useEffectNotOnMount(() => {
+        if (questionTemplateSaveError !== undefined) {
+            setShowCreateErrorMessage(true)
+        }
+    }, [questionTemplateSaveError])
```
denektenina, line 49: Replaced with hook

maiaht (approved):
Looks good!

PR #918, opened 2021-11-29
Upgrade cypress to 9.1

ReedOnly:


ReedOnly:
Issue with running Cypress > 8.4 on Ubuntu-20.04 in Azure CI. This is a know issue with multiple report of issues on both Azure and CircleCI, so I suggest we leave it at version 8.4 for now.

PR #930, opened 2021-12-01
Barrier menu now opens at the correct place

denektenina:
Solution: Since there seemed to be a problem with the way the Button component handled the ref, I wrapped it in a div.

Also moved Menu-code into the component that used it.


denektenina:
Doesn't work

PR #940, opened 2021-12-03
Show PR name on cypress dashboard

maiaht:
We did not manage to figure out how to do this, so this PR will be closed. For more details see the comment below.

### DIFF
--- a/pipelines/cypress-E2E.yml
+++ b/pipelines/cypress-E2E.yml
@@ -59,6 +59,8 @@ jobs:
             displayName: "Start frontend"
 
           - script: |
+                COMMIT_INFO_MESSAGE=$(git log --oneline -2 --pretty=format:%s | tail -1)
+                echo $COMMIT_INFO_MESSAGE
                 npx cypress run \
                 --env FRONTEND_URL=http://localhost:3000,API_URL=http://localhost:5000,AUTH_URL=http://localhost:8080 \
                 --config-file ./cypress.json \


pipelines/cypress-E2E.yml
```diff
@@ -68,5 +68,6 @@ jobs:
                 # avoid warnings about terminal
                 TERM: xterm
                 CYPRESS_RECORD_KEY: $(CYPRESS_RECORD_KEY)
+                COMMIT_INFO_MESSAGE: "This is a test"
```
petterwildhagen, line 71: Will the content for the commit_info_message be extracted from the commit message? Here it is hardcoded

maiaht: This is still a work in progress. This string has been used to test that the content of `COMMIT_INFO_MESSAGE` shows on the Cypress dashboard.

pipelines/cypress-E2E.yml
```diff
@@ -59,6 +59,8 @@ jobs:
             displayName: "Start frontend"
 
           - script: |
+                COMMIT_INFO_MESSAGE=$(git log --oneline -2 --pretty=format:%s | tail -1)
```
maiaht, line 62: The content of line 62 has to be placed under script. The problem is that setting the `COMMIT_INFO_MESSAGE` in the script section does not work. The `COMMIT_INFO_MESSAGE` has to be an environment variable placed under the env section, however, in env we have not managed to set an environment variable to a value that has been set under the script section.

